PR #9, opened 2021-09-06
Fuzzing for validation functions

chris124567:
This PR adds support for fuzzing the transaction, transaction set and header validation functions.  Ran all of them for a while (`go-fuzz -func=FuzzX`) and observed no crashes.

### DIFF
--- a/consensus/validation_fuzz.go
+++ b/consensus/validation_fuzz.go
@@ -1,0 +1,104 @@
+//go:build gofuzz
+// +build gofuzz
+
+package consensus
+
+import (
+	"crypto/ed25519"
+
+	"go.sia.tech/core/types"
+)
+
+func Fuzz(data []byte) int {
+	var txn types.Transaction
+	d := types.NewBufDecoder(data)
+	txn.DecodeFrom(d)
+	if d.Err() != nil {
+		return -1
+	}
+	improveFuzzability(&txn)
+	vc := vcForFuzzing(&txn)
+	if vc.ValidateTransaction(txn) == nil {
+		return 1
+	}
+	return 0
+}
+
+var (
+	anyoneCanSpend        = types.AnyoneCanSpend()
+	anyoneCanSpendAddress = types.PolicyAddress(anyoneCanSpend)
+	zeroPrivkey           = ed25519.NewKeyFromSeed(make([]byte, 32))
+	zeroPubkey            = zeroPrivkey.Public().(ed25519.PublicKey)
+)
+
+func improveFuzzability(txn *types.Transaction) {
+	// TODO: fuzz contract resolutions
+	txn.FileContractResolutions = nil
+
+	// TODO: fuzz foundation address changes
+	txn.NewFoundationAddress = types.VoidAddress
+
+	// set all addresses to AnyoneCanSpend
+	for i := range txn.SiacoinInputs {
+		sci := &txn.SiacoinInputs[i]
+		sci.SpendPolicy = anyoneCanSpend
+		sci.Parent.Address = anyoneCanSpendAddress
+	}
+	for i := range txn.SiafundInputs {
+		sfi := &txn.SiafundInputs[i]
+		sfi.SpendPolicy = anyoneCanSpend
+		sfi.Parent.Address = anyoneCanSpendAddress
+	}
+
+	// override file contract pubkeys so that we can sign them
+	for i := range txn.FileContractRevisions {
+		fcr := &txn.FileContractRevisions[i]
+		copy(fcr.Parent.RenterPublicKey[:], zeroPubkey)
+		copy(fcr.Parent.HostPublicKey[:], zeroPubkey)
+		sigHash := (&ValidationContext{}).ContractSigHash(fcr.Revision)
+		fcr.RenterSignature = types.SignHash(zeroPrivkey, sigHash)
+		fcr.HostSignature = types.SignHash(zeroPrivkey, sigHash)
+	}
+}
+
+func vcForFuzzing(txn *types.Transaction) ValidationContext {
+	var setupTxn types.Transaction
+	for _, sci := range txn.SiacoinInputs {
+		setupTxn.SiacoinOutputs = append(setupTxn.SiacoinOutputs, sci.Parent.SiacoinOutput)
+	}
+	for _, sfi := range txn.SiafundInputs {
+		setupTxn.SiafundOutputs = append(setupTxn.SiafundOutputs, sfi.Parent.SiafundOutput)
+	}
+	for _, fcr := range txn.FileContractRevisions {
+		// ensure that FileContractTax will not overflow when we call ApplyBlock
+		fc := fcr.Parent.FileContract
+		if _, overflow := fc.ValidRenterOutput.Value.AddWithOverflow(fc.ValidHostOutput.Value); overflow {
+			fc.ValidRenterOutput.Value = types.ZeroCurrency
+		}
+		setupTxn.FileContracts = append(setupTxn.FileContracts, fc)
+	}
+	block := types.Block{Transactions: []types.Transaction{setupTxn}}
+
+	au := ApplyBlock(ValidationContext{}, block)
+	au.NewSiacoinElements = au.NewSiacoinElements[1:] // skip block reward
+	for i, sce := range au.NewSiacoinElements {
+		if sce.Address != txn.SiacoinInputs[i].Parent.Address {
+			panic("elements do not match") // sanity check
+		}
+		txn.SiacoinInputs[i].Parent = sce
+	}
+	for i, sfe := range au.NewSiafundElements {
+		if sfe.Address != txn.SiafundInputs[i].Parent.Address {
+			panic("elements do not match") // sanity check
+		}
+		txn.SiafundInputs[i].Parent = sfe
+	}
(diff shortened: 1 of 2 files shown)


chris124567:
Does this look right?  Tried to do the merkle proofs and signatures like you suggested.

I never heard of any fuzzing visualizers. Upon basic inspection there appear to be some papers about designing them but that's not really helpful to us.  It'd be beautiful to have something like `go tool cover` but for go-fuzz (or to make go-fuzz capable of generating output understandable by `go tool cover`).  When mainline Go gets fuzzing there probably will be.

consensus/validation_fuzz.go
```diff
@@ -3,21 +3,54 @@
...
+			Value:   input.Parent.Value,
+			Address: input.Parent.Address,
+		})
+	}
+	for _, input := range txn.SiafundInputs {
+		merkleTxn.SiafundOutputs = append(merkleTxn.SiafundOutputs, types.SiafundOutput{
+			Value:   input.Parent.Value,
+			Address: input.Parent.Address,
+		})
+	}
```
lukechampine, line 31: good start. You'll also need to do this for file contracts.

The `Element` types embed their corresponding `Output` types, so you should be able to write:
```go
	for i, sci := range txn.SiacoinInputs {
		setupTxn.SiacoinOutputs = append(setupTxn.SiacoinOutputs,  sci.Parent.SiacoinOutput)
	}
	for i, sfi := range txn.SiafundInputs {
		setupTxn.SiafundOutputs = append(setupTxn.SiafundOutputs,  sfi.Parent.SiafundOutput)
	}
	for i, fcr := range txn.FileContractRevisions {
		setupTxn.FileContracts = append(setupTxn.FileContracts,  fcr.Parent.FileContract)
	}
	for i, fcr := range txn.FileContractResolutions {
		setupTxn.FileContracts = append(setupTxn.FileContracts,  fcr.Parent.FileContract)
	}
```

You'll also need to sign the contract revisions, and the resolutions will need valid proofs. So, on second thought, maybe for now just set `FileContractResolutions` to `nil` and we'll deal with them later. 😅

consensus/validation_fuzz.go
```diff
@@ -3,21 +3,54 @@
...
+		Transactions: []types.Transaction{merkleTxn},
+	})
+	for i, elem := range au.NewSiacoinElements {
+		txn.SiacoinInputs[i].Parent = elem
+	}
+	for i, elem := range au.NewSiafundElements {
+		txn.SiafundInputs[i].Parent = elem
+	}
+	// use first 32 bytes of transaction data as seed for keys to make it deterministic
+	privkey := ed25519.NewKeyFromSeed(data[:32])
```
lukechampine, line 44: for the signatures to be valid, they need to come from the key that controls the input. So you'll want to derive this key first, then use it to set `input.PublicKey`, and set `output.Address` equal to `types.StandardAddress(pubkey)`.

chris124567:
How's this? I know ordinarily the hoster/renter keys are different but it shouldn't make a difference here.

consensus/validation_fuzz.go
```diff
@@ -0,0 +1,67 @@
...
+	}
+
+	var ctx ValidationContext
+	au := ApplyBlock(ctx, types.Block{
+		Transactions: []types.Transaction{setupTxn},
+	})
+
+	// use first 32 bytes of transaction data as seed for keys to make it deterministic
+	privkey := ed25519.NewKeyFromSeed(data[:32])
+	pubkey := privkey.Public().([ed25519.PublicKeySize]byte)
```
lukechampine, line 38: The underlying type here is `[]byte`, not an array. (In case you're wondering, Go crypto APIs avoid arrays because they are copied by value; don't want multiple copies of key material floating around in memory.)

consensus/validation_fuzz.go
```diff
@@ -0,0 +1,67 @@
...
+		setupTxn.FileContracts = append(setupTxn.FileContracts, fcr.Parent.FileContract)
+	}
+
+	var ctx ValidationContext
+	au := ApplyBlock(ctx, types.Block{
+		Transactions: []types.Transaction{setupTxn},
+	})
+
+	// use first 32 bytes of transaction data as seed for keys to make it deterministic
+	privkey := ed25519.NewKeyFromSeed(data[:32])
```
lukechampine, line 37: probably fine to just use `make([]byte, ed25519.SeedSize)`. This could also panic if `len(data) < 32`.

lukechampine: I suppose this also means that you can make the keys global variables. Would save a bit of CPU during the fuzz call.

consensus/validation_fuzz.go
```diff
@@ -0,0 +1,67 @@
...
+
+	var setupTxn types.Transaction
+	for _, sci := range txn.SiacoinInputs {
+		setupTxn.SiacoinOutputs = append(setupTxn.SiacoinOutputs, sci.Parent.SiacoinOutput)
+	}
+	for _, sfi := range txn.SiafundInputs {
+		setupTxn.SiafundOutputs = append(setupTxn.SiafundOutputs, sfi.Parent.SiafundOutput)
+	}
+	for _, fcr := range txn.FileContractRevisions {
+		setupTxn.FileContracts = append(setupTxn.FileContracts, fcr.Parent.FileContract)
```
lukechampine, line 28: these outputs will need to use `pubkey`:
```go
	for i, sci := range txn.SiacoinInputs {
		sci[i].Parent.Address = types.StandardAddress(pubkey)
 		setupTxn.SiacoinOutputs = append(setupTxn.SiacoinOutputs, sci.Parent.SiacoinOutput)
 	}
```

lukechampine:
I guess if we're just going to use a single keypair for everything, we could just go all the way and use `types.AnyoneCanSpend()`:

```go
var anyoneCanSpendAddr = PolicyAddress(types.AnyoneCanSpend())

// ...

	for i, sci := range txn.SiacoinInputs {
		sci[i].Parent.Address = anyoneCanSpendAddr
 		setupTxn.SiacoinOutputs = append(setupTxn.SiacoinOutputs, sci.Parent.SiacoinOutput)
 	}

// ...

	for i, elem := range au.NewSiacoinElements {
 		txn.SiacoinInputs[i].Parent = elem
 		txn.SiacoinInputs[i].SpendPolicy = types.AnyoneCanSpend()
 	}
```
then you can skip the transaction signing entirely.

The downside, of course, is that the fuzzing won't cover any `SpendPolicy` code. But we can (and probably should) write a separate fuzzer specifically for that.

consensus/validation_fuzz.go
lukechampine: now they need to use `anyoneCanSpendAddress`:
```go
	for i, sci := range txn.SiacoinInputs {
		sci[i].Parent.Address = anyoneCanSpendAddress
 		setupTxn.SiacoinOutputs = append(setupTxn.SiacoinOutputs, sci.Parent.SiacoinOutput)
 	}
```

Basically, we need to overwrite the addresses *before* we call `ApplyBlock`. `ApplyBlock` is going to create a new accumulator that contains the outputs in `setupTxn`; even if you modify `NewSiacoinElements` afterward, the accumulator itself is unaffected (it's just a bunch of hashes). `ValidateTransaction` will use the transaction's Merkle proofs to determine whether each element is in the accumulator or not, so if you modify the elements *after* the accumulator is constructed, the Merkle proofs will become invalid and `ValidateTransaction` will report that the elements aren't in the accumulator.

chris124567: But sci is an element of txn.SiacoinInputs and cannot be indexed

lukechampine: oops, sorry for the confusion there. Should be `txn.SiacoinInputs[i]`

lukechampine:
Ok, I rebased the branch and got the fuzzer to a basically-working state. I didn't actually run `gofuzz`, but I generated a few random txns, ran them through the helper functions I added, and called `ValidateTransaction`. None of the transactions had errors relating to signatures or Merkle proofs -- which is good. Hopefully the fuzzer can get decent coverage now. Can you try running `gofuzz` for a while and see if anything interesting turns up?

chris124567:
Seems good.

`2021/11/04 11:03:01 workers: 4, corpus: 14 (3m47s ago), crashers: 0, restarts: 1/9948, execs: 12912950 (6424/sec), cover: 496, uptime: 33m30s`

lukechampine:
Cool.

My thinking is that the fuzz-related code should live in a separate `fuzz` branch rather than being merged into master. Mostly because I don't like dragging in the gofuzz dependencies; but even when Go 1.18 is out with native fuzzing support, it might be worthwhile to maintain a separate branch. That way, we can do invasive things like directly disabling signature validation (instead of forging the signatures) without polluting `master`. The downside is that `fuzz` would then need to be kept in-sync with `master`, but I suspect this won't be a big issue in practice as long as we don't get too crazy with the specialized fuzz code.

chris124567:
Well I guess in that event we should close this PR without merging to leave the separate branch and just remember to rebase every once in a while.  Unless we want to disable signature verification / that sort of thing in this PR?

lukechampine:
yeah, I'll go ahead and move this code to `fuzz`. Future fuzz-related PRs can target that branch.

PR #11, opened 2021-09-18
Overhaul encoding, port renterhost

chris124567:
This PR ports the `renterhost` modules from lukechampine/us to `core`.  It became necessary to change the encoding system to make types implement encoding/decoding interfaces because something like `WriteRequest(req SiaMarshaler)` is not really possible otherwise.

Other differences to lukechampine/us:
- Changed encoding system to the one in `types`, although `objBuffer` remains
- Removed marshalledSize.  We just encode the value to the `objBuffer`, get the length of the buffer, and then reset the buffer.  This is technically inefficient but since the session `outbuf` is reused over time it shouldn't add that many allocations because while we are writing without a call to `grow` beforehand we would likely just be writing over existing capacity  
- I did not implement any kind of `ReadPrefix` (just calls to `ReadUint64`).  This is obviously a crash vector but I didn't implement it because it seemed it was undecided how to best do it, although the approach in https://github.com/SiaFoundation/core/pull/10 makes the most sense to me

### DIFF
--- a/consensus/accumulator_test.go
+++ b/consensus/accumulator_test.go
@@ -1,14 +1,24 @@
 package consensus
 
 import (
-	"crypto/rand"
-	"encoding/binary"
-	mrand "math/rand"
+	"math"
 	"reflect"
 	"testing"
 
 	"go.sia.tech/core/types"
+	"lukechampine.com/frand"
 )
+
+func randAddr() types.Address {
+	return frand.Entropy256()
+}
+
+func randAmount() types.Currency {
+	return types.NewCurrency(
+		frand.Uint64n(math.MaxUint64),
+		frand.Uint64n(math.MaxUint64),
+	)
+}
 
 func TestBlockRewardValue(t *testing.T) {
 	reward := func(height uint64) types.Currency {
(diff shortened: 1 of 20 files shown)


lukechampine:
Ok, finally found time to go through this. As usual I got kinda sucked in and ended up writing a lot more code than planned...

Most significantly, I decided to reimplement the RHP on top of `mux`. This is basically what RHP3 is anyway, and we don't need to worry about breaking compatibility, so why not? This greatly simplifies the code, because `Session` no longer needs to worry about encryption or framing at all.

I did get a bit lazy with the random values in `session_test.go` -- quite a few fields are simply commented out. Would be nice to restore those.

Also, `gocritic` can get bent lol. I know what I'm doing.

n8maninger:
> we don't need to worry about breaking compatibility

How much backwards compatibility with 1.x are we aiming for?

lukechampine:
I can only think of one property that absolutely must be preserved: addresses. You shouldn't have to tell everyone your "new address" post-hardfork; they should be able to continue sending SC to the address you gave them five years ago. That's why we need `PolicyUnlockConditions`.

But generally speaking, compatibility is going out the window. The transaction format is changing, so the signature format has to change too. And the network protocols. And the file formats. And the HTTP API. etc. Everyone has to upgrade anyway, so there isn't any benefit to preserving e.g. the old p2p protocol. It would be a different story if there were a bunch of third-party implementation of the formats and protocols, but AFAIK there aren't.

The danger is that we get rewrite-happy and try to overhaul *everything* at once. "If we have to break compatibility here *anyway*, I might as well use this opportunity to redesign XYZ..." I'm doing my best to resist that impulse. 😅

lukechampine:
@n8maninger can we disable the failing `gocritic` check (or just disable `gocritic` entirely)?

I know the standard practice for this sort of thing is to add a special comment for the exception or whatever -- but I really don't like muddying up my code with linter-specific magic comments.

PR #12, opened 2021-10-07
Add additional context to errors

n8mgr:
Adds additional context to returned errors

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -19,7 +19,12 @@ jobs:
     timeout-minutes: 10
     steps:
       - uses: actions/checkout@v2
+      - uses: actions/setup-go@v2
+        with:
+          go-version: '1.16'
       - name: test
-        uses: n8maninger/action-golang-test@a8263fc1a85b63937b1b365946518675ceab5740
+        uses: n8maninger/action-golang-test@v1
         with:
           args: "-race"
+          skip-go-install: true
+          show-package-output: true


lukechampine:
👑

net/rhp/session.go
```diff
@@ -133,48 +134,53 @@ func (s *Session) WriteRequest(rpcID Specifier, req ProtocolObject) error {
...
 
 // ReadRequest reads an RPC request using the new loop protocol.
 func (s *Session) ReadRequest(req ProtocolObject, maxLen uint64) (err error) {
-	defer wrapErr(&err, "ReadRequest")
-	return s.readMessage(req, maxLen)
+	err = s.readMessage(req, maxLen)
+	if err != nil {
+		err = fmt.Errorf("ReadRequest: %w", err)
+	}
+	return
```
lukechampine, line 152: could define an `addContext` helper for these callsites as well.

We also don't need to use named returns for these functions anymore.

internal/chainutil/store.go
```diff
@@ -212,22 +223,26 @@ func (fs *FlatStore) Flush() error {
 func (fs *FlatStore) recoverBest(tip types.ChainIndex) error {
 	// if the store is empty, wipe the bestFile too
 	if len(fs.offsets) == 0 {
-		return fs.bestFile.Truncate(0)
+		err := fs.bestFile.Truncate(0)
+		if err != nil {
+			err = fmt.Errorf("failed to truncate best file: %w", err)
+		}
+		return err
```
lukechampine, line 230: preferred style for this is:
```go
if err := fs.bestFile.Truncate(0); err != nil {
    return fmt.Errorf("failed to truncate best file: %w", err)
}
return nil
```
not gonna worry about it though

lukechampine:
hmm, I guess we should test on every version above (and including) the one in our `go.mod`. Our test suite runs pretty fast.

PR #15, opened 2021-10-17
Add Gateway Protocol

chris124567:
Added module similar to `net/rhp` that instead uses the gateway handshake and `net/mux`.  Refactored some of the similarities between rhp and the gateway into `net/rpc`.

### DIFF
--- a/net/mux/mux.go
+++ b/net/mux/mux.go
@@ -210,6 +210,12 @@ func (m *Mux) readLoop() {
 
 // Close closes the underlying net.Conn.
 func (m *Mux) Close() error {
+	// if there's a buffered Write, wait for it to be sent
+	m.mu.Lock()
+	for m.write.header.id != 0 && m.err == nil {
+		m.write.cond.Wait()
+	}
+	m.mu.Unlock()
 	err := m.setErr(ErrClosedConn)
 	if err == ErrClosedConn || err == ErrPeerClosedConn {
 		err = nil
@@ -410,6 +416,10 @@ func (s *Stream) consumeFrame(h frameHeader, payload []byte) {
 func (s *Stream) Read(p []byte) (int, error) {
 	s.cond.L.Lock()
 	defer s.cond.L.Unlock()
+	if !s.established {
+		// developer error: peer doesn't know this Stream exists yet
+		panic("mux: Read called before Write on newly-Dialed Stream")
+	}
 	if !s.rd.IsZero() {
 		if !time.Now().Before(s.rd) {
 			return 0, os.ErrDeadlineExceeded


net/rhp/session_test.go
```diff
@@ -14,10 +14,11 @@ import (
 
 	"lukechampine.com/frand"
 
+	"go.sia.tech/core/net/rpc"
```
lukechampine, line 17: looks like this package wasn't added? I don't understand why this didn't cause a CI failure 🤔

net/gateway/gateway.go
```diff
@@ -0,0 +1,157 @@
...
+		}
+	}
+}
+
+func (s *Session) writeMessage(obj rpc.ProtocolObject) error {
+	if err := s.PrematureCloseErr(); err != nil {
+		return err
+	}
+
+	stream, err := s.mux.DialStream()
```
lukechampine, line 66: we want a 1-to-1 mapping between mux streams and RPCs; the flow for an RPC is:
1. create a new mux stream
2. send rpc ID
3. send rpc message body
4. read rpc response
5. close mux stream

So I think the right approach is to move the `DialStream` into `WriteMessage`, and have `writeMessage` operate on the stream.

net/gateway/gateway.go
```diff
@@ -0,0 +1,157 @@
...
+	if err := s.writeMessage(&rpc.RPCResponse{re, resp}); err != nil {
+		return fmt.Errorf("WriteResponse: %w", err)
+	}
+	return nil
+}
+
+// ReadResponse reads an RPC response. If the response is an error, it is
+// returned directly.
+func (s *Session) ReadResponse(resp rpc.ProtocolObject, maxLen uint64) error {
+	rr := rpc.RPCResponse{nil, resp}
```
lukechampine, line 144: not sure how I feel about a common RPC package (I'll wait until I see it), but assuming we go that route, we'll want to rename `RPCError` -> `Error` and `RPCResponse` -> `Response`

lukechampine: hmm, or maybe put the `rpc` code in the top-level `net`, and keep the prefixes? 🤔 That feels more appropriate.

net/gateway/netaddress.go
```diff
@@ -0,0 +1,166 @@
...
+)
+
+// MaxEncodedNetAddressLength is the maximum length of a NetAddress encoded
+// with the encode package. 266 was chosen because the maximum length for the
+// hostname is 254 + 1 for the separating colon + 5 for the port + 8 byte
+// string length prefix.
+const MaxEncodedNetAddressLength = 266
+
+// A NetAddress contains the information needed to contact a peer.
+type NetAddress string
```
lukechampine, line 17: nah we're not reusing any of this code lol. Just use `string` for now, and we'll deal with the `IsLocal`/`IsLoopback`/etc. stuff later when we implement that actual gateway.

types/encoding.go
```diff
@@ -66,6 +66,12 @@ func (e *Encoder) WritePrefix(i int) { e.WriteUint64(uint64(i)) }
 // WriteTime writes a time.Time value to the underlying stream.
 func (e *Encoder) WriteTime(t time.Time) { e.WriteUint64(uint64(t.Unix())) }
 
+// WriteString writes a string value to the underlying stream.
+func (e *Encoder) WriteString(s string) {
+	e.WritePrefix(len(s))
+	e.Write([]byte(s))
+}
```
lukechampine, line 73: 👍🏻

net/gateway/connect.go
```diff
@@ -0,0 +1,234 @@
...
+	// listener / tcp connections not implemented yet
+	dummyAddr = "127.0.0.1:1"
+)
+
+type sessionHeader struct {
+	publicKey  ed25519.PublicKey
+	genesisID  types.BlockID
+	uniqueID   GatewayID
+	netAddress NetAddress
+}
```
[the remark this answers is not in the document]
lukechampine: (In general, remember that we're writing "dumb" library code here; it should be making very few decisions internally. Let the caller worry about those decisions.)

net/gateway/connect.go
```diff
@@ -0,0 +1,234 @@
...
+		genesisID:  types.GenesisBlock.ID(),
+		uniqueID:   g.uniqueID,
+		netAddress: g.netAddress,
+	}
+	if _, err := exchangeRemoteHeader(conn, ourHeader); err != nil {
+		return nil, err
+	}
+	if err := exchangeOurHeader(conn, ourHeader); err != nil {
+		return nil, err
+	}
```
lukechampine, line 99: idk if it's necessary for these to all be separate functions like they were in the old codebase. Let's try inlining them all.

net/gateway/gateway.go
n8maninger: It's an additional string handshake:
https://gitlab.com/NebulousLabs/siamux/-/blob/master/siamux.go#L301
https://github.com/SiaFoundation/siad/blob/master/modules/host/network.go#L250

https://github.com/lukechampine/us/pull/178

chris124567:
Tried to only do one DialStream call in WriteRequest so we just reuse the stream but this strangely seems to make the subsequent ReadRequest call hang.  I'll keep trying.

Also

>IIRC we only needed to exchange keys so that we could encrypt the connection. mux handles that for us now, so we can ditch this field.

But don't we need their public key when we call dial?

net/gateway/gateway.go
```diff
@@ -0,0 +1,157 @@
+package gateway
+
+import (
+	"fmt"
+	"go.sia.tech/core/net"
+	"go.sia.tech/core/net/mux"
+	"go.sia.tech/core/types"
+	"io"
+	gnet "net"
```
lukechampine, line 9: ehh yeah having to rename a stdlib package kinda sucks. Might be ok if it's rare to import both `net` packages, I guess? But this is definitely evidence against my suggestion.

net/gateway/connect.go
```diff
@@ -0,0 +1,217 @@
...
+	acceptResponse       = "accept"
+	// listener / tcp connections not implemented yet
+	dummyAddr = "127.0.0.1:1"
+)
+
+type SessionConfig struct {
+	publicKey  ed25519.PublicKey
+	genesisID  types.BlockID
+	uniqueID   GatewayID
+	netAddress string
```
lukechampine, line 30: these fields will need to be exported as well.
`GatewayID` can just be `[8]byte`, I think a separate type is overkill.

PR #16, opened 2021-10-18
Extra CI

n8maninger:
Tests multiple versions and operating systems.

Go's CLI tooling is pretty terrible for CI and there's no great way to support benchmark regression, but if we want it I can build something.

### DIFF
--- a/net/mux/errors_windows.go
+++ b/net/mux/errors_windows.go
@@ -1,0 +1,18 @@
+package mux
+
+import (
+	"errors"
+	"io"
+	"syscall"
+)
+
+// isConnCloseError returns true if the error is from the peer closing the
+// connection early.
+func isConnCloseError(err error) bool {
+	return errors.Is(err, io.EOF) ||
+		errors.Is(err, syscall.ECONNRESET) ||
+		errors.Is(err, syscall.EPIPE) ||
+		errors.Is(err, syscall.EPROTOTYPE) ||
+		errors.Is(err, syscall.ECONNABORTED) ||
+		errors.Is(err, syscall.WSAECONNRESET)
+}
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -15,13 +15,17 @@ jobs:
       - name: golangci-lint
         uses: golangci/golangci-lint-action@v2
   test:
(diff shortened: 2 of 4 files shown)


lukechampine:
Looks like the windows test is failing because the syscall error isn't one we expect. Not difficult to fix, although it may be a sign that this test is too brittle.

net/mux/errors_windows.go
```diff
@@ -0,0 +1,12 @@
...
+import (
+	"errors"
+	"syscall"
+)
+
+// isPeerClosedConn returns true if the error is from the peer closing the
+// connection early.
+func isPeerClosedConnErr(err error) bool {
+	return errors.Is(err, syscall.WSAECONNRESET)
+}
```
lukechampine, line 12: hmm, having to add two extra build-tagged files makes me wince a little, but the only alternative I see is:
```go
const WSAECONNRESET syscall.Errno = 10054
if errors.Is(err, io.EOF) ||
	errors.Is(err, syscall.ECONNRESET) ||
	errors.Is(err, syscall.EPIPE) ||
	errors.Is(err, syscall.EPROTOTYPE) ||
	errors.Is(err, syscall.ECONNABORTED) ||
	runtime.GOOS == "windows" && errors.Is(err, WSAECONNRESET) {
	err = ErrPeerClosedConn
}
```
which is obviously fairly gross as well.

I think using build tags is the best option here, but let's move the whole check into the function:

```go
// net/mux/errors.go

func isConnCloseErr(err error) bool {
	return errors.Is(err, io.EOF) ||
		errors.Is(err, syscall.ECONNRESET) ||
		errors.Is(err, syscall.EPIPE) ||
		errors.Is(err, syscall.EPROTOTYPE) ||
		errors.Is(err, syscall.ECONNABORTED)
}


// net/mux/errors_windows.go

func isConnCloseErr(err error) bool {
	return errors.Is(err, io.EOF) ||
		errors.Is(err, syscall.ECONNRESET) ||
		errors.Is(err, syscall.WSAECONNRESET) ||
		errors.Is(err, syscall.EPIPE) ||
		errors.Is(err, syscall.EPROTOTYPE) ||
		errors.Is(err, syscall.ECONNABORTED)
}
```

PR #18, opened 2021-10-26
merkle: expand test coverage

chris124567:
This PR expands test coverage in merkle to about 80%.  Still need to add coverage relating to RevertBlock but that should probably bring it around 90.

### DIFF
--- a/merkle/accumulator_test.go
+++ b/merkle/accumulator_test.go
@@ -8,7 +8,7 @@ import (
 	"lukechampine.com/frand"
 )
 
-func TestUpdateLeaves(t *testing.T) {
+func TestUpdateLeavesSiacoin(t *testing.T) {
 	outputs := make([]types.SiacoinElement, 8)
 	leaves := make([]ElementLeaf, len(outputs))
 	for i := range outputs {
@@ -18,6 +18,11 @@ func TestUpdateLeaves(t *testing.T) {
 	acc.addLeaves(leaves)
 	for i := range outputs {
 		outputs[i].StateElement = leaves[i].StateElement
+	}
+	for _, leaf := range leaves {
+		if !acc.containsLeaf(leaf) {
+			t.Fatal("accumulator missing leaf that was added to it")
+		}
 	}
 
 	updated := []ElementLeaf{
@@ -45,8 +50,14 @@ func TestUpdateLeaves(t *testing.T) {
 		switch i {
 		case 0, 2, 3, 5, 6:
 			addOutput(o, true)
+			if !acc.ContainsSpentSiacoinElement(o) {
+				t.Fatal("accumulator missing spent siacoin element")
+			}
 		default:
 			addOutput(o, false)
+			if acc.ContainsSpentSiacoinElement(o) {
+				t.Fatal("accumulator missing unspent siacoin element")
+			}
 		}
 	}
 	for i := range acc2.Trees {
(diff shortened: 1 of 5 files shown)


merkle/accumulator_test.go
```diff
@@ -61,6 +72,143 @@ func TestUpdateLeaves(t *testing.T) {
...
+		i := 0
+		for ; acc2.hasTreeAtHeight(i); i++ {
+			root = NodeHash(acc2.Trees[i], root)
+		}
+		acc2.Trees[i] = root
+		acc2.NumLeaves++
+	}
+	for _, o := range outputs {
+		addOutput(o, true)
+	}
```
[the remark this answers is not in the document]
lukechampine: When you add another block, existing proofs needs to be updated:
```go
var acc HistoryAccumulator
hau0 := acc.ApplyBlock(blocks[0])
block0Proof := hau0.HistoryProof()
if !acc.Contains(blocks[0], block0Proof) {
    t.Fatal("block 0 not there")
}

hau1 := acc.ApplyBlock(blocks[1])
block1Proof := hau1.HistoryProof()
if !acc.Contains(blocks[1], block1Proof) {
    t.Fatal("block 1 not there after block 1 added")
}
hau1.UpdateProof(&block0Proof)
if !acc.Contains(blocks[0], block0Proof) {
    t.Fatal("block 0 not there after block 1 added")
}
```
...oh, `UpdateProof` doesn't exist. Let me add that real quick...

chris124567:
Added RevertBlock support to the tests but it doesn't seem to be working in either case despite updating the proofs.  In the history accumulator test I revert blocks off the top which seems like the logical thing to do and in the regular accumulator I just revert the `updated` leaves.

lukechampine:
ok, I got `TestApplyBlock` working. It turned out to be quite a bit hairier than I expected, mostly because `ApplyBlock` modifies both `acc` and `updated`. This is annoying. But the alternative (allocating + returning new values) is annoying too, for different reasons. :/
Anyway, it's working now; just a bit more coverage to go. I think we can hit the tricky sections by calling `ApplyBlock` with added leaves multiple times.

chris124567:
Alright, I think this should be good.

merkle/accumulator_test.go
```diff
@@ -155,7 +155,9 @@ func TestApplyBlock(t *testing.T) {
 		leaves = append(leaves, FileContractLeaf(fces[i], false))
 	}
 	var acc ElementAccumulator
-	acc.ApplyBlock(nil, leaves)
+	for i := 0; i < 5; i++ {
+		acc.ApplyBlock(nil, leaves)
+	}
```
lukechampine, line 160: oh dear. I don't even *know* what happens if you add the same leaves multiple times. I think `ApplyBlock` assumes that the leaves have no index/proof, and fills them in -- so this is probably equivalent to adding 4*len(leaves) arbitrary leaves to the accumulator, then calling `ApplyBlock` once on the actual leaves.

To get the coverage we want, I think we can actually just set `acc.NumLeaves` directly; a value like 6 or 9 should work well.

merkle/accumulator_test.go
```diff
@@ -194,8 +196,10 @@ func TestApplyBlock(t *testing.T) {
 	for i := range oldSpent {
 		oldSpent[i].MerkleProof = append([]types.Hash256(nil), oldSpent[i].MerkleProof...)
 	}
-	eau := acc.ApplyBlock(spent, nil)
-
+	var eau ElementApplyUpdate
+	for i := 0; i < 5; i++ {
+		eau = acc.ApplyBlock(spent, nil)
+	}
```
lukechampine, line 202: this `ApplyBlock` is just updating existing leaves, not adding new ones, so I *think* calling it more than once will have no effect on the accumulator (but we might want to confirm that, just in case)

chris124567:
OK I think this should be ready to merge now.

Issue #17, opened 2021-10-25
Spitballing: Attestations


lukechampine:
An *attestation* is a generic key-value object, paired with a pubkey and signature:

```go
type Attestation struct {
    PublicKey PublicKey
    Key       string
    Value     []byte
    Signature Signature
}
```

The general idea is to use attestations for most of the things that currently use `ArbitraryData`. For example, host announcements: instead of encoding a pubkey, IP address, and signature into `ArbitraryData`, you create an attestation with `key = "host announcement"` and `value = [ip address]`. Another example is file contract recovery: create an attestation with `key = "contract backup"` and `value = [host pubkeys + contract IDs + contract keys]`.

So, what's interesting about this? Is this just about adding explicit structure to `ArbitraryData`? No -- it's about making these attestations "first class citizens" of Sia, by *requiring them to be tracked in a separate accumulator*. Each attestation in a transaction is added to this accumulator; existing attestations can also be updated by including a proof for the old leaf. If we then include this accumulator in the block header `Commitment`, then attestations become *verifiable*. That is, you can ask a public explorer,"what's the most-recently-announced IP address for host XYZ?" and the explorer can give you a Merkle proof that the correct answer is ABC.

But this is much cooler when you apply it to personal backups. You can download your most recent contract backup from an explorer without having to trust the explorer. You could also backup your *wallet*, by creating an attestation that contains a list of your spendable outputs. Putting these together allows you to recover your entire Sia "identity" with a single key, in seconds.

To be clear, these use cases are achievable with Sia today. The problem is that they require you to scan the entire blockchain. Attestations allow you to instead request the data from public explorers, with no loss of security. As long as you know the current block hash, you can verify any attestation claimed by the explorer.

 Some concerns:
- Attestations consume space on the blockchain, so we certainly don't want every wallet to be including a backup attestation in every transaction it makes. But this applies to `ArbitraryData` as well; attestations don't change the existing incentives. If necessary, we could increase the "weight" of attestations, meaning clients would have to pay higher fees to include them in transactions. (EDIT: clients could also "compress" their attestations by merely storing the Skynet hash of the data, rather than the data itself.)
[...]

PR #19, opened 2021-11-15
types,consensus: Implement Attestations

lukechampine:
This implements the "dumb" version of attestations; basically just better-structured ArbitraryData. See #17 for discussion.

In this PR, all attestations are validated by consensus. It hadn't occurred to me until just now that perhaps attestations should be validated "lazily;" that is, the consumer of the attestation (e.g. a renter looking for host announcements) should be responsible for validating it. The rationale here is that verifying signatures is slow, so if you don't care about a particular attestation, why should you have to spend CPU cycles validating it?

A counterargument is that enforcing valid signatures in consensus is "safe by default;" no one can accidentally forget to check the signature. Equivalently, invalid signatures are prevented from ever making it on-chain in the first place, which is a nice invariant.

Given that attestations will be comparatively uncommon (evidence: on the current chain, there are ~25x more TransactionSignatures than there are ArbitraryData entries), I'm leaning towards enforcing signatures in consensus. But I'm interested to hear other people's thoughts on this.

### DIFF
--- a/consensus/validation.go
+++ b/consensus/validation.go
@@ -144,6 +144,7 @@ func (vc *ValidationContext) TransactionWeight(txn types.Transaction) uint64 {
 		signatures += len(in.Signatures)
 	}
 	signatures += 2 * len(txn.FileContractRevisions)
+	signatures += len(txn.Attestations)
 
 	return uint64(storage) + 100*uint64(signatures)
 }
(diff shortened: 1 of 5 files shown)


lukehmcc:
Considering the relatively small cost of verifying all signatures (taking into account how little arbitrary data there is on chain as of now) I don't really see a reason to not check them by default. The biggest issue with syncing the chain right now is storage (which ik is being worked on), not CPU power. So unless people get really out of control by dumping a bunch of Attestations on chain to screw with people's sync time, which would be quite an expensive attack, or a new application based on them was built on them I can't foresee it becoming an issue.

n8maninger:
Does `len(attestations)` need to be added to signatures in `ValidationContext.TransactionWeight()`?

consensus/validation.go
```diff
@@ -395,6 +406,15 @@ func (vc *ValidationContext) validFileContractResolutions(txn types.Transaction)
 	return nil
 }
 
+func (vc *ValidationContext) validAttestations(txn types.Transaction) error {
+	for i, a := range txn.Attestations {
+		if !a.PublicKey.VerifyHash(vc.AttestationSigHash(a), a.Signature) {
```
n8maninger, line 411: Should we prevent empty keys and values on attestations as well?

lukechampine: hmm, probably. I can imagine a use case for empty `Data`, but not for empty `Key`.

PR #20, opened 2021-11-19
Improve chainutil test coverage

chris124567:
Currently much of chainutil's coverage comes indirectly from other modules like `chain`.  This PR adds more direct coverage to chainutil.  It raises the coverage from 50% to a little over 75%.  The bulk of the remaining coverage is IO error checking code in store.go which would be pretty difficult to test (at least elegantly).


lukechampine:
thanks!

PR #23, opened 2021-11-25
[WIP] SiaMux frame batching

chris124567:
Working on the preliminary things for this (updating version number, adding message header support) but am running into chacha message authenticaition problems with the payload of the frames.  I think it has to do with the encryptFrame part.  I find the in place encryption/encoding to be kind of difficult to wrap your head around and I think that's where this problem is coming but obviously still this is the part of the code that should be the most performant.


chris124567:
No longer relevant due to https://github.com/SiaFoundation/core/pull/68

PR #24, opened 2021-11-30
rhp: initial renter-host protocol and MDM implementation

n8maninger:
+ Adds renter RHP protocol
+ Adds host RHP protocol
+ Adds host interfaces from `us`
+ Adds ephemeral account interface
+ Adds initial implementation of MDM
+ Adds host registry interface
+ Adds more of the `us ` Merkle package

Needs polish and some implementations aren't completed, but is this a direction we're okay with going?


net/rhp/objects.go
```diff
@@ -586,3 +580,430 @@ func (r *RPCWriteResponse) DecodeFrom(d *types.Decoder) {
...
+
+// RPCSettingsResponse contains the JSON-encoded settings for a host.
+type RPCSettingsResponse struct {
+	Settings []byte
+}
+
+// MaxLen returns the maximum encoded length of an object. Implements rpc.Object.
+func (r *RPCSettingsResponse) MaxLen() int {
+	// JSON encoded settings should result in < 2 KiB
+	return 3000
```
lukechampine, line 592: might as well use `defaultMaxLen`

host/rpc.go
```diff
@@ -0,0 +1,213 @@
...
+	buf, err := json.Marshal(settings)
+	if err != nil {
+		sh.log.Errorln("failed to marshal settings:", err)
+		return
+	}
+
+	// write the price table to the stream. The price table is sent
+	// before payment so the renter can determine if they want to
+	// continue interacting with the host.
+	err = rpc.WriteResponse(stream, &rhp.RPCSettingsResponse{
```
lukechampine, line 179: something feels weird here. The host sends its settings to the renter, then waits for the renter to pay it, but if the renter just disconnects without paying, it's fine?

Did the original code send the prices first, then wait for payment, then send the full settings?

net/rhp/objects.go
```diff
@@ -586,3 +580,430 @@ func (r *RPCWriteResponse) DecodeFrom(d *types.Decoder) {
...
+
+// MaxLen returns the maximum encoded length of an object. Implements rpc.Object.
+func (req *RPCExecuteProgramRequest) MaxLen() int {
+	return defaultMaxLen
+}
+
+// EncodeTo encodes a RPCExecuteProgramRequest to an encoder. Implements types.EncoderTo.
+func (req *RPCExecuteProgramRequest) EncodeTo(e *types.Encoder) {
+	req.FileContractID.EncodeTo(e)
+	e.WriteUint64(uint64(len(req.Instructions)))
```
lukechampine, line 640: `WritePrefix`

net/rhp/objects.go
```diff
@@ -586,3 +580,430 @@ func (r *RPCWriteResponse) DecodeFrom(d *types.Decoder) {
...
+		instr.EncodeTo(e)
+	}
+	e.WriteUint64(req.ProgramDataLength)
+}
+
+// DecodeFrom decodes a RPCExecuteProgramRequest from a decoder. Implements
+// types.DecoderFrom.
+func (req *RPCExecuteProgramRequest) DecodeFrom(d *types.Decoder) {
+	req.FileContractID.DecodeFrom(d)
+	req.Instructions = make([]Instruction, d.ReadUint64())
```
lukechampine, line 651: `ReadPrefix`

net/rhp/objects.go
```diff
@@ -586,3 +580,430 @@ func (r *RPCWriteResponse) DecodeFrom(d *types.Decoder) {
...
+	PayByEphemeralAccount = rpc.NewSpecifier("PayByEphemAcc")
+)
+
+// WithdrawalMessage is the amount of money to deduct from the account to create
+// the RPC budget.
+type WithdrawalMessage struct {
+	AccountID types.PublicKey
+	Expiry    uint64
+	Amount    types.Currency
+	Nonce     [8]byte
```
lukechampine, line 670: this is the sort of field that I'd want a comment on -- why does a withdrawal need a nonce?

net/rhp/objects.go
```diff
@@ -586,3 +580,430 @@ func (r *RPCWriteResponse) DecodeFrom(d *types.Decoder) {
...
+
+// PayByContractRequest is a request to create an RPC budget using funds from a
+// file contract.
+type PayByContractRequest struct {
+	ContractID            types.ElementID
+	NewRevisionNumber     uint64
+	NewValidRenterPayout  types.Currency
+	NewValidHostPayout    types.Currency
+	NewMissedRenterPayout types.Currency
+	NewMissedHostPayout   types.Currency
```
lukechampine, line 729: `rhp.ContractOutputs`

net/rhp/objects.go
```diff
@@ -586,3 +580,430 @@ func (r *RPCWriteResponse) DecodeFrom(d *types.Decoder) {
...
+	return 32 + 32 + 16 + 8
+}
+
+// EncodeTo encodes a Receipt to an encoder. Implements
+// types.EncoderTo.
+func (r *Receipt) EncodeTo(e *types.Encoder) {
+	r.Host.EncodeTo(e)
+	r.Account.EncodeTo(e)
+	r.Amount.EncodeTo(e)
+	e.WriteUint64(uint64(r.Timestamp.Unix()))
```
lukechampine, line 871: `WriteTime`

net/rhp/objects.go
```diff
@@ -586,3 +580,430 @@ func (r *RPCWriteResponse) DecodeFrom(d *types.Decoder) {
...
+	e.WriteUint64(resp.NewDataSize)
+	e.WriteUint64(uint64(len(resp.Proof)))
+	for _, h := range resp.Proof {
+		h.EncodeTo(e)
+	}
+	if resp.Error == nil {
+		e.WriteUint64(0)
+	} else {
+		e.WriteString(resp.Error.Error())
+	}
```
lukechampine, line 948:
```go
var errStr string
if resp.Error != nil {
    errStr = resp.Error.Error()
}
e.WriteString(errStr)
```

host/programexecutor.go
```diff
@@ -0,0 +1,665 @@
...
+		}
+
+		// read the sector data.
+		var sector [rhp.SectorSize]byte
+		if _, err = io.ReadFull(r, sector[:]); err != nil {
+			err = fmt.Errorf("failed to read append sector data: %w", err)
+			break
+		}
+
+		proof, err = ex.executeAppendSector(&sector, instruction.Arguments[8] == 1)
```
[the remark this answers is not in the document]
lukechampine: oh I see, they're currently ignored. Will have to check with the Skynet team

net/rhp/program.go
```diff
@@ -0,0 +1,226 @@
...
+
+// Instruction is a single instruction in a program.
+type Instruction struct {
+	Specifier [16]byte
+	Arguments []byte
+}
+
+// MaxLen implements rpc.Object
+func (i *Instruction) MaxLen() int {
+	return 272
```
lukechampine, line 73: magic?

net/rhp/program.go
```diff
@@ -0,0 +1,226 @@
...
+	// AdditionalCollateral cost is the additional collateral the host should add
+	// during program finalization
+	AdditionalCollateral types.Currency
+
+	Memory uint64
+	Time   uint64
+}
+
+// Add adds the usage from b to r, returning r.
+func (r *ResourceUsage) Add(b ResourceUsage) *ResourceUsage {
```
lukechampine, line 108: should be a value receiver

net/rhp/program.go
```diff
@@ -0,0 +1,226 @@
...
+	costs.BaseCost = settings.ProgInitBaseCost.Add(resourceCost(settings, data, time))
+	costs.Memory = 1 << 20
+	return
+}
+
+// finalizeCost returns the cost of finalizing a program.
+// note: siad's finalize cost uses the program's total memory usage. Since memory
+// cost is already included in the base cost of the instruction, I removed it
+// to avoid double charging.
+func finalizeCost(settings types.HostSettings) (costs ResourceUsage) {
```
lukechampine, line 145: `finalize` is a verb -- should be `finalizationCost`

net/rhp/program.go
```diff
@@ -0,0 +1,226 @@
...
+	// note: in siad the additional collateral does not consider remaining
+	// duration.
+	costs.AdditionalCollateral = settings.Collateral.Mul64(SectorSize * duration)
+	return
+}
+
+// DropSectorsCost returns the cost of the drop sectors instruction.
+func DropSectorsCost(settings types.HostSettings, n uint64) (costs ResourceUsage) {
+	costs.BaseCost = settings.InstrDropSectorsUnitCost.Mul64(n).
+		Add(settings.InstrDropSectorsBaseCost)
```
lukechampine, line 184: unwrap

renter/executeprogram.go
```diff
@@ -0,0 +1,158 @@
...
+)
+
+// ExecuteProgram executes the program on the renter. execute is called for each
+// successfully executed instruction; the reader passed to execute is only
+// valid until the next call. Returning an error terminates remote program
+// execution. The program must be fully executed for any changes to be committed.
+// If an instruction fails, the program is rolled back.
+//
+// ContractRevision and RenterKey may be nil if the program is not read-write.
+func (s *Session) ExecuteProgram(program Program, input *bytes.Buffer, payment PaymentMethod, execute ExecuteFunc) error {
```
lukechampine, line 42: `bytes.Buffer` is a weird argument; why not `[]byte`?

renter/formcontract.go
```diff
@@ -0,0 +1,157 @@
...
+		return types.FileContractRevision{}, nil, errors.New("host revision signature is invalid")
+	}
+
+	for i := range hostSigs.ContractSignatures {
+		txn.SiacoinInputs[i].Signatures = append(txn.SiacoinInputs[i].Signatures, hostSigs.ContractSignatures[i]...)
+	}
+
+	txns = append(txns, resp.Parents...)
+	txns = append(txns, txn)
+	return
```
lukechampine, line 156: `return fcr, append(resp.Parents, txn), nil`

renter/payment.go
```diff
@@ -0,0 +1,146 @@
...
+	"go.sia.tech/core/net/rhp"
+	"go.sia.tech/core/net/rpc"
+	"go.sia.tech/core/types"
+	"lukechampine.com/frand"
+)
+
+type (
+	// A PaymentMethod pays for RPC usage during an renter-host session.
+	PaymentMethod interface {
+		Pay(s *mux.Stream, amount types.Currency) error
```
lukechampine, line 18: this is the sort of thing I was talking about earlier -- feels wrong to abstract stream side-effects via an interface. This should be `Pay(amount types.Currency) (something, error)`

renter/payment.go
```diff
@@ -0,0 +1,146 @@
...
+		expiry    uint64
+	}
+
+	payByContract struct {
+		rev             *types.FileContractRevision
+		secret          ed25519.PrivateKey
+		hostKey         types.PublicKey
+		refundAccountID types.PublicKey
+
+		manager ChainManager
```
lukechampine, line 33: `cm` is the conventional name for a `ChainManager`

renter/programbuilder.go
```diff
@@ -0,0 +1,269 @@
...
+func (pb *ProgramBuilder) AddAppendSectorInstruction(sector *[rhp.SectorSize]byte, proof bool) *ProgramBuilder {
+	if pb.err != nil {
+		return pb
+	}
+
+	args := make([]byte, rhp.LenAppendSectorArgs)
+	binary.LittleEndian.PutUint64(args, uint64(pb.data.Len()))
+	if proof {
+		args[8] = 1
+	}
```
lukechampine, line 58:
```go
var args bytes.Buffer
e := types.NewEncoder(&args)
args.WriteUint64(uint64(pb.data.Len()))
args.WriteBool(proof)
e.Flush()
```

renter/programbuilder.go
```diff
@@ -0,0 +1,269 @@
...
+
+func (pb *ProgramBuilder) setErr(err error) error {
+	if pb.err == nil {
+		pb.err = err
+	}
+	return pb.err
+}
+
+// AddAppendSectorInstruction adds an append sector instruction to the program.
+func (pb *ProgramBuilder) AddAppendSectorInstruction(sector *[rhp.SectorSize]byte, proof bool) *ProgramBuilder {
```
lukechampine, line 49: "builder" objects returning themselves so that you can chain methods together is a JavaScript-ism. No real need for it here.

renter/session.go
```diff
@@ -0,0 +1,219 @@
...
+)
+
+// A Session is an implementation of the renter side of the renter-host protocol.
+type Session struct {
+	secret  ed25519.PrivateKey
+	hostKey types.PublicKey
+
+	wallet  Wallet
+	tpool   TransactionPool
+	manager ChainManager
```
lukechampine, line 54: `cm`

renter/session.go
```diff
@@ -0,0 +1,219 @@
...
+		Address() types.Address
+		Addresses() []types.Address
+		FundTransaction(txn *types.Transaction, amount types.Currency, pool []types.Transaction) ([]types.ElementID, func(), error)
+		SignTransaction(vc consensus.ValidationContext, txn *types.Transaction, toSign []types.ElementID) error
+	}
+)
+
+// A Session is an implementation of the renter side of the renter-host protocol.
+type Session struct {
+	secret  ed25519.PrivateKey
```
lukechampine, line 49: `privkey`

renter/session.go
```diff
@@ -0,0 +1,219 @@
...
+	hostKey types.PublicKey
+
+	wallet  Wallet
+	tpool   TransactionPool
+	manager ChainManager
+	session *rhp.Session
+
+	settingsMu         sync.RWMutex
+	settings           types.HostSettings
```
lukechampine, line 58: why do the settings need to be protected by mutex? Are they updated in parallel with other RPCs?

lukechampine: also, default to `sync.Mutex` -- it often outperforms `RWMutex`, even on read-heavy workloads. If it's a bottleneck, we can always swap it later.

renter/session.go
```diff
@@ -0,0 +1,219 @@
...
+	settingsExpiration time.Time
+}
+
+func (s *Session) currentSettings() (types.HostSettings, error) {
+	s.settingsMu.RLock()
+	defer s.settingsMu.RUnlock()
+	if s.settingsExpiration.Before(time.Now()) {
```
lukechampine, line 65: the eternal question: which is more readable, `expiration.Before(time.Now())`, or `time.Now().After(expiration)`? :P

renter/session.go
```diff
@@ -0,0 +1,219 @@
...
+	if err := rpc.ReadResponse(stream, &resp); err != nil {
+		return types.ZeroCurrency, fmt.Errorf("failed to read account balance response: %w", err)
+	}
+
+	return resp.Balance, nil
+}
+
+// Settings returns the current settings for the host. If a payment method is
+// provided the setting UID should be used when interacting with the host.
+func (s *Session) Settings(payment PaymentMethod) (settings types.HostSettings, _ error) {
```
lukechampine, line 158: consider defining a `NoPayment` type that can be passed instead of `nil`

renter/session.go
```diff
@@ -0,0 +1,219 @@
...
+		return types.HostSettings{}, fmt.Errorf("failed to pay for settings: %w", err)
+	}
+
+	if err = rpc.ReadResponse(stream, &rhp.RPCTrackingResponse{}); err != nil {
+		return types.HostSettings{}, fmt.Errorf("failed to read tracking response: %w", err)
+	}
+
+	s.settingsMu.Lock()
+	s.settings = settings
+	s.settingsExpiration = time.Now().Add(settings.Validity)
```
lukechampine, line 191: computing `time.Now().Add(settings.Validity)` should be done as soon as possible after receiving the settings

renter/session.go
```diff
@@ -0,0 +1,219 @@
...
+// NewSession initiates a new RHP session with the host.
+func NewSession(netaddress string, theirKey types.PublicKey, ourKey ed25519.PrivateKey, wallet Wallet, tpool TransactionPool, manager ChainManager) (*Session, error) {
+	conn, err := net.Dial("tcp", netaddress)
+	if err != nil {
+		return nil, fmt.Errorf("failed to open connection: %w", err)
+	}
+
+	s := &Session{
+		hostKey: theirKey,
+		secret:  ourKey,
```
lukechampine, line 206: what is this key being used for?

lukechampine (changes requested):
phew!

host/host.go
```diff
@@ -0,0 +1,386 @@
...
+	sh.contractLocks[id].waiters--
+	if sh.contractLocks[id].waiters <= 0 {
+		delete(sh.contractLocks, id)
+	}
+}
+
+// lockRegistryKey locks the registry key with the provided key preventing
+// updates. The context can be used to interrupt if the registry key lock cannot
+// be acquired quickly.
+func (sh *SessionHandler) lockRegistryKey(ctx context.Context, key types.Hash256) error {
```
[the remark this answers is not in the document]
n8maninger: We keep the contract validation logic outside of the contract store, does it makes sense to similarly keep that logic outside of the registry store?

host/host.go
```diff
@@ -0,0 +1,386 @@
...
+	sh.contractMu.Lock()
+	defer sh.contractMu.Unlock()
+	if _, exists := sh.contractLocks[id]; !exists {
+		return
+	}
+	sh.contractLocks[id].c <- struct{}{}
+	sh.contractLocks[id].waiters--
+	if sh.contractLocks[id].waiters <= 0 {
+		delete(sh.contractLocks, id)
+	}
```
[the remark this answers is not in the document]
n8maninger: I started with a `sync.Cond`, but in this context, waking up all waiters with `Broadcast` to check the lock has never sat right with me.

host/program.go
```diff
@@ -0,0 +1,219 @@
...
+	}
+
+	settings, err := sh.validSettings(settingsUID)
+	if err != nil {
+		sh.log.Warnf("settings uid %s not found:", settingsUID)
+		rpc.WriteResponse(stream, nil, err)
+		return
+	}
+
+	budget, err := sh.processPayment(stream)
```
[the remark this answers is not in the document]
n8maninger: Contract payments need to write the host signature back to the stream. Having a helper with direct access to the stream seems cleaner and keeps everything isolated.

host/program.go
```diff
@@ -0,0 +1,219 @@
...
+	if err != nil {
+		sh.log.Warnln("execute program:", "failed to process payment:", err)
+		rpc.WriteResponse(stream, nil, err)
+		return
+	}
+	defer budget.Refund()
+
+	// wrap the stream in a budget to pay for bandwidth usage, also serves as
+	// a data limit since all usage is deducted from the stream budget.
+	budgetedStream := newBudgetDataLimit(stream, budget, settings)
```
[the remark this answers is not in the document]
n8maninger: Deducting from the budget directly seems cleaner than tracking the number of bytes written/read then deducting at the end. Simple to short-circuit the RPC if the budget runs out due to execution or bandwidth costs.

host/program.go
```diff
@@ -0,0 +1,219 @@
...
+	// read the program
+	var executeReq rhp.RPCExecuteProgramRequest
+	err = rpc.ReadResponse(budgetedStream, &executeReq)
+	if err != nil {
+		sh.log.Warnln("execute program:", "failed to read execute program request:", err)
+		rpc.WriteResponse(stream, nil, errors.New("failed to read execute request"))
+		return
+	}
+
+	var requiresContract, readwrite bool
```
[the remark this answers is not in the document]
n8maninger: Seems backward, to me, to always check the negative `!readonly` to expect finalization instead of the positive `readwrite`

host/host.go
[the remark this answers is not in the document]
lukechampine: it's not ideal, but neither is allocating a new channel for each lock ¯\\\_(ツ)\_/¯
At some point we'll benchmark it and see how perf scales with number of waiters. And try to measure how many waiters the average host can expect at any given time...

host/programexecutor.go
```diff
@@ -0,0 +1,665 @@
...
+
+	err := ex.payForExecution(rhp.SwapSectorCost(ex.settings))
+	if err != nil {
+		return nil, fmt.Errorf("failed to pay instruction cost: %w", err)
+	}
+
+	// swap the sector roots.
+	ex.newRoots[indexA], ex.newRoots[indexB] = ex.newRoots[indexB], ex.newRoots[indexA]
+	// update the program's contract state
+	ex.newMerkleRoot = merkle.MetaRoot(ex.newRoots)
```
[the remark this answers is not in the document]
n8maninger: Currently, we're writing an `RPCExecuteInstrResponse` followed by the output after every instruction executes. Are you wanting to batch all of the instructions together into a single write at the end?

types/host.go
```diff
@@ -0,0 +1,368 @@
...
+	d.Read(r.Signature[:])
+}
+
+// RegistryKey is the unique key for a RegistryValue.
+func RegistryKey(pub PublicKey, tweak Hash256) Hash256 {
+	h := hasherPool.Get().(*Hasher)
+	defer hasherPool.Put(h)
+	h.Reset()
+
+	encodeCompatSiaPublicKey(h.E, pub)
```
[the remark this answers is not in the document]
n8maninger: I think we have to retain compatibility, like we did with `PolicyUnlockConditions`, since there are existing apps built on registry lookups.

types/host.go
```diff
@@ -0,0 +1,368 @@
...
+	InstrRevisionBaseCost Currency `json:"instrRevisionBaseCost"`
+	// InstrWriteBaseCost is the base cost in Hastings to execute a write
+	// instruction.
+	InstrWriteBaseCost Currency `json:"instrWritebasecost"`
+	// InstrReadRegistryBaseCost is the base cost in Hastings to execute a read
+	// registry instruction.
+	InstrReadRegistryBaseCost Currency `json:"instrReadRegistryBaseCost"`
+	// InstrUpdateRegistryBaseCost is the base cost in Hastings to execute an
+	// update registry instruction.
+	InstrUpdateRegistryBaseCost Currency `json:"instrUpdateRegistryBaseCost"`
```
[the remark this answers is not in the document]
n8maninger: Doing `val, ok := settings.Get(rhp.SettingMaxCollateral)` doesn't seem very pleasant, but having a more dynamic settings structure could be nice. We could also consider binary encoding it for `RPCSettingsResponse` instead of JSON.

host/programexecutor.go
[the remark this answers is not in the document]
n8maninger: They seemed redundant; differing instruction lengths invalidate the program regardless. We could pass a program version uint64 instead if we're worried about upgradability or robustness.

host/rpc.go
n8maninger: This follows the siad RHP3 settings RPC, but it only ever returned the price table. I combined the RHP2 external settings and the RHP3 price table into `HostSettings`. 

The renter needs the settings first to know how much to pay for "registering" the settings. The renter can choose not to pay if it decides the prices are too high.

host/host.go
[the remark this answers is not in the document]
lukechampine: yeah, that's fair. Generally, stores should be "dumb" -- they should only be concerned with serialization and ACID. So I guess what I'm arguing is that we should treat the `Registry` not as a store, but as its own thing, *wrapping* a `RegistryStore`.

Maybe that's overkill though? idk. Mostly I just don't like having all these individual locks. For something like the registry, I suspect that a single lock would be perfectly adequate, or perhaps a CAS+retry approach.

host/program.go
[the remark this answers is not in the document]
lukechampine: fair. Let's at least rename it, though. `newBudgetDataLimit` does not suggest that the return value is an `io.Reader`.

host/program.go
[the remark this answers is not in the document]
lukechampine: maybe `writeable`, then? `readwrite` feels like it was directly taken from POSIX perm bits. And it's ill-suited for the cases where you're exclusively writing and not reading anything.

host/programexecutor.go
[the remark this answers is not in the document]
lukechampine: I see. That certainly does simplify the proof code.
There are lots of ways we can make the proof calculation more efficient, so let's not worry about it for now.

types/host.go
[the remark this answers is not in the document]
lukechampine: I imagine you'd only need to check for support once, and after that you can access the field directly:
```go
if _, ok := settings.Prices[rpc.InstrUpdateRegistry]; !ok {
    return errors.New("operation not supported")
}
cost := settings.Prices[rpc.InstrUpdateRegistry].Mul64(foo) // etc.
```

The settings were JSON encoded so that they could be extended later without breaking old renters, but we haven't made much use of that capability. A binary-encoded dynamic map might be a good compromise between efficiency and flexibility.

host/rpc.go
lukechampine: Maybe withhold the settings ID until after payment, then? I don't like writing an empty struct to signal success, feels janky.

I'm not in love with having a settings table in general; host prices don't change that frequently, and when they do, the host can return an "outdated settings" error, the renter can re-request the new settings, and then retry. Whereas with locked-in settings, you have to maintain an arbitrarily-sized map in memory with expiration dates.

n8maninger: I agree I don't think we necessarily need to register the price tables, and we can get by with short-lived settings.

lukechampine: yeah, historically we've had lots of weird bugs where the host would just return "low paying output" or something and it wasn't clear if the duration was wrong, or if the settings changed, or something else. Would definitely like to get to the bottom of that.

host/payment.go
```diff
@@ -0,0 +1,204 @@
...
+// processEAPayment processes a payment using an ephemeral account.
+func (sh *SessionHandler) processEAPayment(stream *mux.Stream) (*rpcBudget, error) {
+	height := sh.cm.Tip().Height
+
+	var req rhp.PayByEphemeralAccountRequest
+	if err := rpc.ReadResponse(stream, &req); err != nil {
+		return nil, fmt.Errorf("failed to read EA payment request: %w", err)
+	}
+
+	// validate that the request is valid.
```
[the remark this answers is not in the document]
n8maninger: Seems to be an accurate comment 😆

host/programexecutor.go
```diff
@@ -0,0 +1,665 @@
...
+
+		proof, err = ex.executeAppendSector(&sector, instruction.Arguments[8] == 1)
+	case rhp.InstrDropSectors:
+		if len(instruction.Arguments) != rhp.LenDropSectorsArgs {
+			err = errInvalidArgs
+			break
+		}
+
+		var dropped uint64
+		if err = binary.Read(r, binary.LittleEndian, &dropped); err != nil {
```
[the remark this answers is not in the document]
n8maninger: Because arguments for drop sectors in the RHP3 MDM is a uint64 data offset to the uint64 number of sectors to drop.

If we're not worried about compatibility, it might be worth ditching the arguments/data distinction and moving to something closer to the RHP2 RPC loop. A loop of 16-byte op-code, uint64 data length, then the data to execute the instruction, a `done` op-code would signal finalization.

host/contracts.go
```diff
@@ -0,0 +1,259 @@
...
+	/*if err := vc.ValidateTransactionSet(txnset); err != nil {
+		sh.log.Warnln("form contract:", "failed to validate transaction set:", err)
+		rpc.WriteResponse(stream, nil, fmt.Errorf("failed to validate transaction set: %w", err))
+		return
+	}*/
+
+	contract := Contract{
+		ID: types.ElementID{
+			Source: types.Hash256(txn.ID()),
+			Index:  0,
```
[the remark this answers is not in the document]
n8maninger: Added helpers for all output types

host/host.go
```diff
@@ -0,0 +1,386 @@
...
+		UnconfirmedParents(txn types.Transaction) ([]types.Transaction, error)
+		BroadcastTransaction(txn types.Transaction, dependsOn []types.Transaction)
+	}
+)
+
+// A Contract is a file contract paired with various metadata.
+type Contract struct {
+	types.FileContractRevision
+
+	ID types.ElementID
```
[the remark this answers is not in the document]
n8maninger: I think it'd be fine to pass the signature for the initial contract without bumping the revision number.

host/program.go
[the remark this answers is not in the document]
n8maninger: I dropped read-only and read-write in favor of `requiresFinalization` and `requiresContract`. More accurate as `UpdateRegistry` would be considered `readonly`.

renter/payment.go
```diff
@@ -0,0 +1,147 @@
...
+		refundAccountID types.PublicKey
+
+		cm ChainManager
+	}
+)
+
+var (
+	// NoPayment used to signal no payment.
+	NoPayment PaymentMethod
+)
```
lukechampine, line 41: Ah, interesting. I had originally meant something like this:
```go
type noPayment struct{}

func (noPayment) Pay(*mux.Stream, types.Currency) error { return nil }

func NoPayment() PaymentMethod { return noPayment{} }
```
That way, it would always be safe to call `payment.Pay` in the renter RPCs without checking.

I see now, though, that this doesn't work, because the RPC's behavior is *supposed* to change depending on whether or not you've indicated that you're paying.

renter/programbuilder.go
```diff
@@ -0,0 +1,192 @@
...
+	"errors"
+
+	"go.sia.tech/core/net/rhp"
+	"go.sia.tech/core/types"
+)
+
+// A ProgramBuilder constructs programs for the renter to execute on a host.
+type ProgramBuilder struct {
+	instructions                           []rhp.Instruction
+	requiresFinalization, requiresContract bool
```
lukechampine, line 15: I would put these on separate lines so that the padding on other fields isn't so extreme

renter/programbuilder.go
```diff
@@ -0,0 +1,192 @@
...
+
+// AddSwapSectorInstruction adds a swap sector instruction to the program.
+func (pb *ProgramBuilder) AddSwapSectorInstruction(i, j uint64, proof bool) {
+	instr := rhp.InstrSwapSector{
+		RootAOffset:   uint64(pb.data.Len()),
+		ProofRequired: proof,
+	}
+	binary.Write(pb.data, binary.LittleEndian, i)
+	instr.RootBOffset = uint64(pb.data.Len())
+	binary.Write(pb.data, binary.LittleEndian, j)
```
lukechampine, line 128: should use an encoder for things like this

renter/session.go
```diff
@@ -0,0 +1,245 @@
...
+
+	if err := json.Unmarshal(resp.Settings, &settings); err != nil {
+		return rhp.HostSettings{}, fmt.Errorf("failed to decode settings: %w", err)
+	}
+
+	if payment == nil {
+		return
+	} else if err := payment.Pay(stream, settings.RPCHostSettingsCost); err != nil {
+		return rhp.HostSettings{}, fmt.Errorf("failed to pay for settings: %w", err)
+	}
```
lukechampine, line 212: maybe we should just have two separate `Session` methods, one for paid settings, one for unpaid

lukechampine: (or two separate RPCs, even!)

types/types.go
```diff
@@ -318,6 +350,23 @@ func (h BlockHeader) ID() BlockID {
...
+}
+
+// FoundationOutputID returns the output ID of the foundation payout. A Foundation subsidy output is only created every
+// 4380 blocks after the hardfork at block 298000.
+func (h BlockHeader) FoundationOutputID() ElementID {
+	return ElementID{
+		Source: Hash256(h.ID()),
+		Index:  1,
+	}
+}
```
lukechampine, line 368: Not sure if I like these being methods of `BlockHeader`... We can revisit it later, I guess.

host/host.go
```diff
@@ -0,0 +1,387 @@
...
+		// into w. Returns the number of bytes read or an error.
+		ReadSector(root types.Hash256, w io.Writer, offset, length uint64) (n uint64, err error)
+	}
+
+	// An EphemeralAccountStore manages ephemeral account balances.
+	EphemeralAccountStore interface {
+		Balance(accountID types.PublicKey) (types.Currency, error)
+		Deposit(accountID types.PublicKey, amount types.Currency) (types.Currency, error)
+		Refund(accountID types.PublicKey, amount types.Currency) error
+		Withdraw(accountID types.PublicKey, requestID types.Hash256, amount types.Currency) error
```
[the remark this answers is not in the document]
n8maninger: `requestID` can be used to uniquely identify a withdrawal and prevent duplicates from succeeding.

consensus/validation.go
```diff
@@ -15,11 +15,13 @@ import (
 )
 
 const (
-	blocksPerDay  = 144
-	blocksPerYear = 144 * 365
+	// BlocksPerDay is the number of blocks in a day, based on the target block time of 10 minutes.
+	BlocksPerDay = 144
+	// BlocksPerYear is the number of blocks in a year, based on the target block time of 10 minutes.
+	BlocksPerYear = 144 * 365
```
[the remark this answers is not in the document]
n8maninger: I exported them for `net/rhp` and `host`. I think we should avoid having multiple definitions of `BlocksPerYear`, but I don't think exporting a second const in `net/rhp` is too bad a compromise.

host/program.go
```diff
@@ -0,0 +1,211 @@
...
+
+	"go.sia.tech/core/net/mux"
+	"go.sia.tech/core/net/rhp"
+	"go.sia.tech/core/net/rpc"
+)
+
+type (
+	// A budgetedStream limits reads and writes to a ReadWriter using an RPC budget. Writes subtract the download
+	// bandwidth price multiplied by the number of bytes written and reads subtract the upload bandwidth price
+	// multiplied by the number of bytes read.
```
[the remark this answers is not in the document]
n8maninger: 😐

host/programexecutor.go
```diff
@@ -0,0 +1,574 @@
...
+			// read the root from the program's data
+			if _, err := io.ReadFull(r, root[:]); err != nil {
+				return nil, fmt.Errorf("failed to read sector root: %w", err)
+			}
+
+			// read the offset and length from the program's data
+			if err := binary.Read(r, binary.LittleEndian, &offset); err != nil {
+				return nil, fmt.Errorf("failed to read sector offset: %w", err)
+			}
+			if err := binary.Read(r, binary.LittleEndian, &length); err != nil {
```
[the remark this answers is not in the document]
n8maninger: Yeah, I've been holding off on finalizing much more of RPCExecute; there are quite a few places we can improve and refactor depending on the final implementation.

net/rhp/program.go
```diff
@@ -0,0 +1,539 @@
...
+}
+
+// HasSectorCost returns the cost of the has sector instruction.
+func HasSectorCost(settings HostSettings) (costs ResourceUsage) {
+	costs.BaseCost = settings.InstrHasSectorBaseCost
+	return
+}
+
+// ReadCost returns the cost of the read instruction.
+func ReadCost(settings HostSettings, l uint64) (costs ResourceUsage) {
```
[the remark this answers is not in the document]
n8maninger: If we don't mind keeping it off the interface type it should be fine, differing parameters is the only reason I've left them standalone.

renter/session.go
n8maninger: Maybe `ScanSettings` and `RegisterSettings`?  The API and implementation are pretty clean, though; I can't think of a good reason to break it up.

n8maninger: We had talked about getting rid of settings registration entirely, which would make the Settings RPC unpaid anyway.

host/programexecutor.go
```diff
@@ -0,0 +1,574 @@
...
+		settings: settings,
+		budget:   budget,
+		duration: 1,
+		host:     sh,
+		vc:       vc,
+
+		gainedSectors:  make(map[types.Hash256]struct{}),
+		removedSectors: make(map[types.Hash256]struct{}),
+	}
+	pe.encoder = types.NewEncoder(&pe.output)
```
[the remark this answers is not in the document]
n8maninger: It's pretty awkward, but I don't think that makes sense since the executor expects exclusive control of its output buffer, and `ExecuteInstruction` requires an `io.Writer` for flushing the response + buffer.

consensus/validation.go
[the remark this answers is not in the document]
lukechampine: meh, it's not exactly a difficult calculation to make. I don't see an issue with having multiple definitions, as long as they're unexported.

renter/programbuilder.go
```diff
@@ -0,0 +1,193 @@
...
+)
+
+// A ProgramBuilder constructs programs for the renter to execute on a host.
+type ProgramBuilder struct {
+	instructions         []rhp.Instruction
+	requiresFinalization bool
+	requiresContract     bool
+	encoder              *types.Encoder
+	data                 *bytes.Buffer
+	n                    uint64
```
lukechampine, line 19: this is a bit too terse; should be named `offset` (I think?)

renter/session.go
lukechampine: yeah, probably don't need a separate RPC, but let's have separate session methods.

types/encoding.go
```diff
@@ -457,7 +457,7 @@ func (e *Encoder) WritePolicy(p SpendPolicy) {
 			}
 			writeUint8(p.SignaturesRequired)
 		default:
-			panic("unhandled policy type")
+			panic(fmt.Sprintf("unhandled policy type %v", p))
```
lukechampine, line 460: use `%T` to print the type rather than the value

lukechampine (commented):
Okay, this is coming together well. Let's narrow the scope a bit by merging some things separately -- in particular, the `blake2b` package, the `merkle` stuff, the changes to `net/rpc`, and the use of `tb.TempDir()`.

host/payment.go
```diff
@@ -0,0 +1,169 @@
...
+
+	"go.sia.tech/core/net/mux"
+	"go.sia.tech/core/net/rhp"
+	"go.sia.tech/core/net/rpc"
+	"go.sia.tech/core/types"
+)
+
+type rpcBudget struct {
+	value types.Currency
+}
```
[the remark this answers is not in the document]
n8maninger: Eh, we'll have to agree to disagree lol... The original implementation made it convenient and clear how the budget is managed. Now, I think you're right. However, removing it altogether would add unnecessary boilerplate back to the RPC. 

We can't calculate how much bandwidth the user will use and deduct it at the start, and deducting it once at the end requires keeping track and dealing with underflows during the RPC anyway. Due to the program's arguments only being offsets, we can't calculate the cost of executing an instruction until we read its data; reading all data into memory to frontload costs is inefficient. Having a helper to simplify budget tracking, underflows, and refunds directly makes the most sense and removes unnecessary boilerplate from the RPC.

renter/payment.go
n8maninger: I think the only other way this would work would be changing the interface to `IsPaymentMethod()`, then switching on the type in a concrete `processPayment`, but that seems worse. The contract payment method requires a back and forth, so I can't just return an `rpc.Object` and write it to the stream.

consensus/validation.go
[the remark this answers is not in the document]
n8maninger: I was thinking that it would be beneficial for everywhere to have a concrete assertion on what `BlocksPerYear` is.

lukechampine:
ok, should rebase this on top of master now

host/programexecutor.go
```diff
@@ -0,0 +1,477 @@
...
+	// TODO: calculate optional proof.
+	return nil, nil
+}
+
+func (pe *programExecutor) executeContractRevision() error {
+	var contract types.FileContractRevision
+	if pe.contract == nil {
+		return errors.New("no contract revision set")
+	}
+	contract.EncodeTo(pe.encoder)
```
lukechampine, line 205: should be `pe.contract`?

host/registry.go
```diff
@@ -0,0 +1,175 @@
...
+
+	// registry entries must be locked while they are being modified
+	registryMu    sync.Mutex
+	registryLocks map[types.Hash256]*locker
+}
+
+// lockRegistryKey locks the registry key with the provided key preventing
+// updates. The context can be used to interrupt if the registry key lock cannot
+// be acquired quickly.
+func (r *registry) lockRegistryKey(key types.Hash256, timeout time.Duration) error {
```
lukechampine, line 37: since these are methods on the `registry` type, we can probably drop `Registry` from their name

lukechampine: ditto for `registryMu` and `registryLocks`

host/rpc.go
```diff
@@ -0,0 +1,258 @@
...
+	log.Infoln("host read settings ID")
+
+	settings, err := sh.validSettings(settingsID)
+	if err != nil {
+		log.Warnf("settings uid %s not found:", settingsID)
+		rpc.WriteResponseErr(stream, err)
+		return
+	}
+
+	log.Infoln("got settings")
```
lukechampine, line 89: debugging?

host/rpc.go
```diff
@@ -0,0 +1,258 @@
...
+			Host:      sh.privkey.PublicKey(),
+			Account:   req.AccountID,
+			Amount:    fundAmount,
+			Timestamp: time.Now(),
+		},
+	}
+
+	h := types.NewHasher()
+	resp.Receipt.EncodeTo(h.E)
+	resp.Signature = sh.privkey.SignHash(h.Sum())
```
lukechampine, line 141: Probably worth defining a `hashObject(obj types.EncoderTo)` helper function. It'll be a bit less efficient, but improved readability makes it a net win imo

net/rhp/program.go
```diff
@@ -0,0 +1,633 @@
...
+	case InstrReadSector:
+		v.EncodeTo(e)
+	case InstrContractRevision:
+		v.EncodeTo(e)
+	case InstrSwapSector:
+		v.EncodeTo(e)
+	case InstrUpdateRegistry:
+		v.EncodeTo(e)
+	case InstrReadRegistry:
+		v.EncodeTo(e)
```
lukechampine, line 583: we could either add `rpc.Object` to the `Instruction` interface, or you could use a type assertion:
```go
et, ok := instr.(types.EncoderTo)
if !ok {
    panic(fmt.Sprintf("%T does not implement types.EncoderTo", v))
}
et.EncodeTo(e)
```

net/rhp/program.go
```diff
@@ -0,0 +1,633 @@
...
+
+func readInstruction(d *types.Decoder) Instruction {
+	var spec rpc.Specifier
+	d.Read(spec[:])
+
+	switch spec {
+	case specInstrAppendSector:
+		var instr InstrAppendSector
+		instr.DecodeFrom(d)
+		return instr
```
lukechampine, line 597: oof. You *could* condense this a bit with a `map[rpc.Specifier]Instruction`, but I think that would end up allocating N new objects per call instead of 1. You can at least do this, though:
```go
var inst Instruction
switch spec {
case specInstrAppendSector:
    instr = new(InstrAppendSector)
}
instr.DecodeFrom(d)
return instr
```
(Note that this would require all `Instruction` methods to take pointer receivers; this is fine and should be done regardless. Using value types isn't much more efficient in this case, because if the value is *ever* converted to an `Instruction` interface, it has to be heap-allocated anyway; whereas if it's a pointer, you don't need a separate allocation to convert to an interface.)

net/rhp/registry.go
```diff
@@ -0,0 +1,109 @@
...
+
+	h := types.NewHasher()
+
+	h.E.Write(r.Tweak[:])
+	h.E.WriteBytes(data)
+	h.E.WriteUint64(r.Revision)
+
+	return types.Work{
+		NumHashes: h.Sum(),
+	}
```
lukechampine, line 70: should be `types.WorkRequiredForHash(types.BlockID(h.Sum())`

renter/formcontract.go
```diff
@@ -0,0 +1,153 @@
...
+	"fmt"
+
+	"go.sia.tech/core/net/rhp"
+	"go.sia.tech/core/net/rpc"
+	"go.sia.tech/core/types"
+)
+
+// FormContract negotiates a new contract with the host using the specified
+// funds and duration.
+func (s *Session) FormContract(renterKey types.PrivateKey, hostFunds, renterFunds types.Currency, endHeight uint64) (types.FileContractRevision, []types.Transaction, error) {
```
lukechampine, line 14: I think this function should just return a `FileContract`, i.e. the `Revision` field of what it's returning now. Having a whole `FileContractRevision` isn't really useful to the renter at this stage. (And it won't be valid anyway, since it doesn't have an accumulator proof.)

host/contracts.go
```diff
@@ -0,0 +1,312 @@
...
+		txn.SiacoinInputs[i].Signatures = append(txn.SiacoinInputs[i].Signatures, renterSigs.ContractSignatures[i]...)
+	}
+
+	txnset := append(parents, txn)
+
+	/*if err := vc.ValidateTransactionSet(txnset); err != nil {
+		log.Warnln("form contract:", "failed to validate transaction set:", err)
+		rpc.WriteResponse(stream, nil, fmt.Errorf("failed to validate transaction set: %w", err))
+		return
+	}*/
```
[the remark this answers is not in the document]
n8maninger: It should be, but we can't use `vc.ValidateTransactionSet` with the existing stub interfaces in the tests, I'll need to replace the existing stubs with a real validation context and hot wallet.

renter/formcontract.go
n8maninger: I chose to return a `FileContractRevision` to pass the signatures to the caller. Should we consider moving the signatures directly to the `FileContract`?

host/host.go
```diff
@@ -0,0 +1,356 @@
...
+			rpcFn(stream)
+		}()
+	}
+}
+
+// NewSessionHandler initializes a new host session manager.
+func NewSessionHandler(privkey types.PrivateKey, cm ChainManager, ss SectorStore, cs ContractStore, as EphemeralAccountStore, rs RegistryStore, w Wallet, sr SettingsReporter, tp TransactionPool, log Logger) *SessionHandler {
+	h := types.NewHasher()
+	privkey.PublicKey().EncodeTo(h.E)
+	hostID := h.Sum()
```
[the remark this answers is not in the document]
n8maninger: With v1 or?

host/contracts.go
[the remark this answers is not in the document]
n8maninger: I couldn't decide if I should add a full-on `ghost` style package or not.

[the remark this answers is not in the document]
lukechampine: hmm. I think we should strive to make our tests match real world conditions where possible, so I'm in favor of constructing + validating an actual transaction here (rather than e.g. defining and stubbing a `TransactionValidator` interface).

host/host.go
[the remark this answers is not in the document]
lukechampine: yeah, I think v1 hashes pubkeys differently than v2 -- v1 pubkeys have a specifier, whereas v2 pubkeys are always assumed to be ed25519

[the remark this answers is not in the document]
n8maninger: Oh, you're right. We'll need to use a v1 compatible encoding for the registry's host id.

lukechampine:
Why do we need a `walletutil.EphemeralWallet` and a `walletutil.TxPool`? Can't we reuse the existing implementations?

n8maninger:
> Why do we need a `walletutil.EphemeralWallet` and a `walletutil.TxPool`? Can't we reuse the existing implementations?

Do you mean the stubs I had in the renter package? There was no real implementation on them. I thought separating the logic into its own package made more sense.

lukechampine:
Oh, I see the problem -- I was thinking of the `wallet` and `txpool` packages, but those are in `siad`, not `core`. Hmm. So I guess our options are:

1. Duplicate the implementations
2. Move `wallet` and `txpool` to `core`
3. Move `host` to `siad`

I'm torn between these; they all have downsides. What are your thoughts?

n8maninger:
I would be more inclined to move `wallet` and `txpool` to `core` over moving `host` to `siad`, but I don't see a problem with a less complete testing implementation in `core`. My thought is that all host and renter implementations should share the same base communication layer, which fits `core`. That isn't necessarily the case for `wallet` or `txpool`.

lukechampine:
IMO `core` is for types, consensus code, and protocols/formats. A good heuristic is: "Will someone ever want to write a different implementation of this (in Go)?" If the answer is yes, it probably doesn't belong in `core`. For example, there are lots of ways you could design a Sia wallet, so it wouldn't make sense to have one "blessed" wallet in `core`. But all of those wallets will compute their signatures the same way, so the code for computing signatures should live in `core`. Similarly, there are many ways to design a Sia renter; but all of them will speak the RHP, so the RHP code should live in `core`.

But the `host` package is kind of ambiguous. By design, it imposes very few "choices" of its own; it mostly just glues various interfaces together, with some validation and convenience types thrown in. In that sense, it feels `core`-like: it handles all the boring/subtle stuff that you don't want to worry about when implementing a host. On the other hand, `host` is not *completely* "pure;" it's far more likely that someone would write a competing `host` package than e.g. a competing `merkle` or `types` package. So does it pass the heuristic, or not?

`p2p` is in the same boat, if you think about it. We could certainly define a `p2p.Server` in `core` that listens for connections, spawns a request loop, passes request objects to some sort of `Handler` interface, etc. And unlike `host`, we wouldn't need like ten different interfaces and tons of glue code. Does *this* pass the heuristic?

The asymmetry here is kinda surprising. When you design a *client* for a protocol (e.g. the renter), the correct approach is obvious: just provide a bunch of building blocks, and let the user worry about combining them. But *servers* are not so obvious -- just look at `net/http`. I'm not totally sure why. I wonder, if we forgot about connection lifecycles, spawning goroutines, etc., and just provided "server building blocks," what would the `host` package look like?

n8maninger:
If we only want to provide server building blocks, I don't think we need anything more than what `net/siamux`, `net/rhp`, and `net/rpc` already provide. We could add a `net/rpc` server that accepts an RPC handler and spawns goroutines similar to `http.Server`, but I'm not sure how beneficial that would be on top of siamux.

lukechampine:
Oh, I mean considering the host as a whole as a "server." So accepting connections and spawning handlers is a building block, but more important building blocks would be standalone functions like `validContractFormation`, and types like `budget`, `registryManager `, `programExecutor`, etc.

It occurs to me that the presence of a logger is probably another indication that something doesn't belong in `core`, lol. It's pretty much impossible to introduce logging in a way that will always fit everyone's usecase.

lukechampine:
To summarize, we would divide this code between `core` and `siad` as follows:

`core`:
- Specifiers and request/response types for the RHP and MDM
- `HostSettings`
- Most validation functions
- Most interfaces
- `rpcBudget` and `budgetedStream`
- `programExecutor`
- `registryManager`, `RegistryValue`, etc.
- Ephemeral implementations of interfaces (for testing)
- A generic RPC server, to be used by both `host` and `p2p`
- `renter.Contract`
- `renter.Program` and `renter.ProgramBuilder`
- `renter.PaymentMethod` and its implementations (maybe?)

`siad`:
- `contractManager`
- `Logger` interface
- `locker`
- Interface implementations (for testing and for production)
- Most `SessionHandler` code (e.g. `handleRPCFormContract`)
- Most `renter.Session` code

Does that sound reasonable?

PR #26, opened 2021-12-15
merkle: add blake2b

n8maninger:
Adds the `us` blake2b package


merkle/stack.go
```diff
@@ -0,0 +1,172 @@
+package merkle
```
lukechampine, line 1: I think this code only pertains to the RHP; let's put it in `net/rhp`

lukechampine: should change `merkle.NodeHash` to be a passthrough to `blake2b.SumPair`, though.

lukechampine:
Merged in https://github.com/SiaFoundation/core/commit/17613a9a4a3cbc03df21a5d74ea1c64e162d2356

PR #33, opened 2022-01-11
ci: add nightly test workflow

n8maninger:
Runs the test suite nightly with -count=1000 to try and detect NDFs

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -1,0 +1,26 @@
+name: Nightly
+
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: '0 0 * * *'
+
+jobs:
+  test:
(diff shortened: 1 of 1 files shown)


.github/workflows/nightly.yml
```diff
@@ -0,0 +1,28 @@
...
+    timeout-minutes: 120
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-go@v2
+        with:
+          go-version: ${{ matrix.go-version }}
+      - name: test
+        uses: n8maninger/action-golang-test@v1
+        with:
+          args: "-race;-count=1000;-timeout=30m"
```
lukechampine, line 26: same as `siad`, I would use `-count=100` and `-failfast`

PR #34, opened 2022-01-19
Add renter/host interfaces and logic

n8maninger:
Moves most protocol logic into the `siad` repo keeps base interfaces and shared logic. We can move pieces into separate PRs as they're finalized.

+ MDM implementation
+ Registry implementation
+ Interfaces for data stores
+ Contract helpers


renterhost/mdm/mdm.go
```diff
@@ -0,0 +1,553 @@
...
+	time := 1 + instructions
+	costs.BaseCost = settings.ProgInitBaseCost.Add(resourceCost(settings, data, time))
+	costs.Memory = 1 << 20
+	return
+}
+
+// finalizationCost returns the cost of finalizing a program. note: siad's
+// finalize cost uses the program's total memory usage. Since memory cost is
+// already included in the base cost of the instruction, I removed it to avoid
+// double charging.
```
lukechampine, line 472: should probably avoid first-person pronouns in docstring

renterhost/mdm/mdm.go
```diff
@@ -0,0 +1,553 @@
...
+	costs.BaseCost = resourceCost(settings, costs.Memory, costs.Time)
+	return
+}
+
+// ExecutionCost returns the cost of initializing and, optionally, finalizing a
+// program.
+func ExecutionCost(settings renterhost.HostSettings, data, instructions uint64, requiresFinalization bool) (costs ResourceUsage) {
+	costs = initCost(settings, data, instructions)
+	if !requiresFinalization {
+		costs = costs.Add(finalizationCost(settings))
```
lukechampine, line 485: flipped boolean?

renterhost/registry/registry.go
```diff
@@ -0,0 +1,187 @@
...
+	// ed25519 specifier + LE uint64 pub key length + public key
+	buf := make([]byte, 16+8+32)
+	copy(buf, "ed25519")
+	binary.LittleEndian.PutUint64(buf[16:], 32)
+	copy(buf[24:], pub[:])
+	return types.HashBytes(buf)
+}
+
+// ValidEntry validates the fields of a registry entry.
+func ValidEntry(value Value) (err error) {
```
lukechampine, line 127: my rule of thumb for naming here is:

- `ValidX` returns a `bool`
- `ValidateX` returns an `error`

types/types.go
```diff
@@ -499,6 +550,15 @@ func HashBytes(b []byte) Hash256 { return blake2b.Sum256(b) }
 // implementation whose constructor returns a concrete type.
 var hasherPool = &sync.Pool{New: func() interface{} { return NewHasher() }}
 
+// HashObject computes the hash of the object using Sia's hash function.
+func HashObject(obj EncoderTo) Hash256 {
```
lukechampine, line 554: is this being used anywhere? It forces an allocation and usually we're hashing multiple objects together, not just one

lukechampine (commented):
looks pretty good -- I do wonder if we need separate packages for `rhp` and `renterhost` though

renterhost/contracts.go
```diff
@@ -0,0 +1,274 @@
...
+
+// ProgramRevision returns a new file contract revision with the burn amount
+// subtracted from the missed host output.
+func (c *Contract) ProgramRevision(burn types.Currency) (types.FileContract, error) {
+	if c.Revision.MissedHostOutput.Value.Cmp(burn) < 0 {
+		return types.FileContract{}, errors.New("not enough funds")
+	}
+	fcr := c.Revision
+	fcr.RevisionNumber++
+	fcr.MissedHostOutput.Value = fcr.MissedHostOutput.Value.Sub(burn)
```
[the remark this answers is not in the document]
n8maninger: Instead of having a burn output, the missed proof outputs can't be greater than the sum of the valid proof outputs and the sum of the valid proof outputs must be the same as the parent contract.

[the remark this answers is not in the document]
lukechampine: oh yeah. Forgot I did that lol

types/types.go
n8maninger: Yes, ephemeral accounts hash and sign the withdrawal request

n8maninger:
> looks pretty good -- I do wonder if we need separate packages for `rhp` and `renterhost` though

A split between the netcode and protocol objects in `net/rhp` and validation logic into a top-level `renterhost` made sense, but I don't feel particularly strongly either way.

lukechampine:
Hmm. I can see the rationale, but ultimately I don't think the packages are different enough to outweigh the downsides of having both a `renterhost` and an `rhp` package. We can always split out new packages later without breaking compatibility, but we can't remove an existing one. In the same vein, I'd probably consolidate the `mdm` and `registry` packages; they're not very big, and they don't have any consumers that wouldn't also be importing `net/rhp` anyway. It's ok for `net/rhp` to be a large package; it's a large protocol.

lukechampine:
ok, structure looks good to me. Any blockers keeping this a WIP, or are we good to merge?

n8maninger:
I think we're good, but we'll probably have a few more iterations. I'll push up the siad side of this PR.

PR #38, opened 2022-01-26
consensus: use parent when reverting revisions

n8maninger:
A block revert should ideally return the prior revision instead of returning the reverted revision

### DIFF
--- a/consensus/update.go
+++ b/consensus/update.go
@@ -88,7 +88,7 @@ func applyHeader(vc *ValidationContext, h types.BlockHeader) {
 	vc.Index = h.Index()
 }
 
-func updatedInBlock(vc ValidationContext, b types.Block) (scos []types.SiacoinElement, sfos []types.SiafundElement, revised, resolved []types.FileContractElement, leaves []merkle.ElementLeaf) {
+func updatedInBlock(vc ValidationContext, b types.Block, apply bool) (scos []types.SiacoinElement, sfos []types.SiafundElement, revised, resolved []types.FileContractElement, leaves []merkle.ElementLeaf) {
 	addLeaf := func(l merkle.ElementLeaf) {
 		// copy proofs so we don't mutate transaction data
 		l.MerkleProof = append([]types.Hash256(nil), l.MerkleProof...)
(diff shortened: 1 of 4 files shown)


consensus/update.go
```diff
@@ -108,7 +108,9 @@ func updatedInBlock(vc ValidationContext, b types.Block) (scos []types.SiacoinEl
 		}
 		for _, fcr := range txn.FileContractRevisions {
 			fce := fcr.Parent
-			fce.FileContract = fcr.Revision
+			if !revert {
+				fce.FileContract = fcr.Revision
+			}
```
lukechampine, line 113: This will change the result of `merkle.FileContractLeaf`, so I suspect that this will cause `(merkle.ElementAccumulator).RevertBlock` to produce an invalid `RevertUpdate`: specifically, if you use it to update an existing element proof, then the proof won't match the original (that is, applying-then-reverting will result in a different proof than not-applying-in-the-first-place). Would be good to confirm this in a test.

We could get around this by leaving `FileContractLeaf` alone, and only touching the element appended to `resolved` or `revised`. I'm not totally sold on this change yet though.

lukechampine:
I assume the motivation here is that we want to write something like this:
```go
func (h *Host) ProcessChainRevertUpdate(cru *RevertUpdate) error {
    for _, fce := range cru.RevisedFileContracts {
        if _, ok := h.contracts[fce.ID]; ok {
            h.contracts[fce.ID]
        }
    }
}
```
Note that it's already possible to achieve this by iterating over the block's transactions, rather than the update -- but it's not particularly intuitive that you'd need to do that. It seems natural that the `RevertUpdate` should reflect the state of the chain prior to the block, not after it. Whereas currently, the `SpentSiacoins`, `ResolvedFileContracts`, etc. fields in both `ApplyUpdate` and `RevertUpdate` reflect the effect of applying the block.

There's something to be said for this design, but frankly when I wrote it I just didn't know exactly what we'd want out of those types yet. Now that we have more sophisticated modules subscribing to these updates, it seems prudent to revisit the design. In particular, instead of having `RevisedFileContracts` and `ResolvedFileContracts` fields, we could have `UpdateFileContract` and a `FileContractWasResolved` methods.

n8maninger:
Yes, that was the motivation. The contractor currently has to loop through the transactions looking for reverted revisions. 

The current API for reverts is unintuitive, but it works for other modules. The wallet store can add `SpentSiacoinOutputs` back into its unspent map. The transaction pool adds all transactions in the block back to the pool. It would follow that revisions should also be exposed in a valid state.

lukechampine:
ok, I refactored `updatedInBlock` to set the `spent` bool properly (in addition to the revision) depending on whether we're applying or reverting.

Another reason to use methods, rather than exposing fields, is that you can prevent clients from accidentally aliasing memory (e.g. not calling `DeepCopy` on a transaction, or not copying the Merkle proof of an element). I'll take a stab at a refactor shortly.

PR #39, opened 2022-01-30
Add ReadSector method

n8maninger:
Add a `ReadSector` method to the `rhp` package, modify the program executor to use it, and update the host interfaces with additional changes.

### DIFF
--- a/net/rhp/rpc.go
+++ b/net/rhp/rpc.go
@@ -746,6 +746,14 @@ type WithdrawalMessage struct {
 	Nonce [8]byte
 }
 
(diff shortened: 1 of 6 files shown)


net/rhp/rpc.go
```diff
@@ -746,6 +746,14 @@ type WithdrawalMessage struct {
 	Nonce [8]byte
 }
 
+// Hash computes the hash of the withdrawal message used for signing the
+// pay by ephemeral account request.
+func (wm *WithdrawalMessage) Hash() types.Hash256 {
```
lukechampine, line 751: let's call this `SigHash`, for parity with other signed hashes

PR #41, opened 2022-02-01
mux: add DialStreamContext

n8maninger:


lukechampine:
Hmm, I think this test is failing nondeterministically on CI

n8maninger:
> expected error context deadline exceeded, got read: peer closed underlying connection

It seems to fail every time on the CI, but I can't reproduce it locally...  It looks like it's caused by the error handling in the echo server closing the connection; the server should probably ignore stream errors since we're testing short-circuiting it.

n8maninger:
It looks like we're also getting intermittent connection errors in both `TestContext` and `TestManyStreams`
```
--- FAIL: TestManyStreams (42.26s)
[18](https://github.com/n8maninger/coretest/runs/5490340056?check_suite_focus=true#step:4:18)
    mux_test.go:105: accept mux error: version handshake failed: could not read peer version: EOF
[19](https://github.com/n8maninger/coretest/runs/5490340056?check_suite_focus=true#step:4:19)
    mux_test.go:132: version handshake failed: could not read peer version: read tcp 127.0.0.1:60007->127.0.0.1:60005: wsarecv: An existing connection was forcibly closed by the remote host.
```
```
version handshake failed: could not read peer version: EOF
```

PR #42, opened 2022-02-01
Add contract revision helpers

n8maninger:
Separates the contract revision helpers from the contract type.

### DIFF
--- a/net/rhp/contracts.go
+++ b/net/rhp/contracts.go
@@ -56,42 +56,40 @@ func (c *Contract) MaxLen() uint64 {
 
 // PaymentRevision returns a new file contract revision with the amount added to
 // the host payout fields and subtracted from the renter payout fields.
-func (c *Contract) PaymentRevision(amount types.Currency) (types.FileContract, error) {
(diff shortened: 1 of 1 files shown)


net/rhp/contracts.go
```diff
@@ -69,26 +70,26 @@ func (c *Contract) PaymentRevision(amount types.Currency) (types.FileContract, e
...
-	fcr.MissedRenterOutput.Value = fcr.ValidRenterOutput.Value
-	return fcr
+func ClearingRevision(fc types.FileContract) (rev types.FileContract) {
+	rev = fc
+	rev.RevisionNumber = math.MaxUint64
+	rev.Filesize = 0
+	rev.FileMerkleRoot = types.Hash256{}
+	rev.MissedHostOutput.Value = rev.ValidHostOutput.Value
+	rev.MissedRenterOutput.Value = rev.ValidRenterOutput.Value
+	return
```
lukechampine, line 83: need to copy the whole output, not just the value. Also, I'd probably just reuse `fc` instead of introducing a new variable

PR #43, opened 2022-02-01
Add validation for ephemeral Siafund and contract revision outputs

n8maninger:


lukechampine:
This made me pause for a minute to think about maturity delays. Clearly, if an output requires a maturity delay, you won't be able to use it as an ephemeral output!

Sia v1 adds a 144-block maturity delay to various outputs:
1. Block reward outputs
2. Foundation subsidy outputs
3. Siafund claim outputs
4. Storage proof outputs (both valid and missed)

The reasoning for 1 and 2 follows that of Bitcoin: these outputs create new coins, and their IDs are derived from the block ID that creates them (rather than from a previous transaction). If such outputs could be spent immediately, then reorgs would frequently cause problems, because any transactions "tainted" by those outputs would become invalid. Normally, when a reorg causes a transaction to be reverted, the transaction will return to the txpool and can be included in a subsequent block. Even if you have a chain of dependent transactions A->B->C, you can still "restore" things by reincluding all of them. But transactions derived from a reverted block can *never* be reincluded. That's why it's important to put some "distance" between coin creation and coin use.

I should note here that this is *not* an issue of security of correctness. It's a UX issue! Sia and Bitcoin could work even with no maturity delay at all. But people don't like it when their transactions are invalidated, and without a maturity delay, a reorg could invalidate lots of transactions. Imagine if you immediately sent a block reward output to a big exchange, and the exchange bundled that output into a giant transaction that executed a bunch of trades -- all of those trades would be "tainted" and become invalid after a reorg, causing a massive headache and lots of unhappy users.
(I should also note that some blockchains do not have a maturity delay, notably Ethereum. Zcash and Grin have also considered removing the maturity delay, though AFAIK they still have one. Their argument is that the burden is on users to avoid immature outputs. I'm not convinced.)

So, how does this reasoning extend to case 3? Well, the value of a siafund claim output is determined by the current value of the "siafund pool." Say you make a transaction A that spends 1 SF, creating a claim output of 10 SC. Then you use the 10 SC output as the input to transaction B. But then a reorg occurs, reverting both A and B. You rebroadcast A, and it works fine; but when you try to rebroadcast B, it fails! It fails because the reorg changed the value of the siafund pool, and thus changed the value of the claim output; maybe it's only worth 7 SC now. So it's clear that claim outputs should have a maturity delay to prevent this situation.
Interestingly, a maturity delay is *not* applied to regular siafund outputs. Siafund outputs have a `ClaimStart` field that records the value of the siafund pool when the output was created, so you might assume that a maturity delay is warranted here. But this field is only used when computing the value of a subsequent claim output, and claim outputs already have a maturity delay; so the `ClaimStart` value effectively has a delay attached to it by proxy.

Lastly, storage proofs. There are two factors at play here. First, the entropy required for a (valid) storage proof is derived from the block immediately prior to the contract's `WindowStart`. Clearly, a reorg can change this block, causing the proof to become invalid -- so valid storage proof outputs need a maturity delay. Second, if a proof is not submitted before the `WindowEnd` height, then the proof is deemed "missed" and the contract's missed outputs are created. So if we reorg to a lower height, then these outputs will not exist, and there is no transaction that we can immediately rebroadcast to restore them; we just have to wait until the height is above `WindowEnd` again. (We can also observe that reorging to a *higher* height, above `WindowEnd`, could cause a valid storage proof to become invalid -- another reason to add a maturity delay to valid proofs.)

In Sia v2, storage proofs work a little differently. Missed proofs do not automatically trigger output creation; instead, the contract must be explicitly resolved as missed by broadcasting a transaction with a `FileContractResolution`. The net effect is the same, though: such transactions may become invalid if a reorg occurs, so a maturity delay is still required.

Sia v2 also has [early contract resolution](https://github.com/SiaFoundation/core/pull/31), allowing a contract's outputs to be released early if a special revision is applied. So: do we need a maturity delay here? Unfortunately, if we're being strict about it, I think we do. Yes, unlike normal contract resolution, we can immediately rebroadcast the revision transaction -- but that's not even to save us if the chain reorgs *above* the contract's `WindowEnd`. Still, I think there's a chance that we could "get away with" removing the maturity delay on early resolution, specifically if we are confident that the benefits outweigh the risks.

Overall, I was surprised to discover how murky this question is. Even among experts, people either disagree on the merits of the maturity height, or don't have a clear understanding of what exactly it accomplishes. I should probably seek the advice of people smarter than myself before making a decision here. For now, I propose that we add maturity delays for siafund claim outputs and file contract resolution outputs. Consequently, this PR becomes moot -- sorry. :/

n8maninger:
Added timelocks to contract resolution outputs in #44

lukechampine:
Further thought: *Any* output of a transaction containing a contract revision or resolution should have a maturity delay. Revisions and resolutions can become (permanently) invalidated by reorg, so any outputs deriving from those transactions will be invalidated as well.

A more aggressive option would be to ban such transactions from having outputs at all; any contracts would have to be funded using precisely-value inputs. v1 renters and hosts already do this (via a setup transaction), which is probably why v1 has never suffered from this issue. It's not a consensus rule, however.

The problem with setup transactions it that they're wasteful. Instead of forming a contract in 1 transaction, it takes 3. So I'd prefer to take the "all outputs are subject to a maturity delay" approach instead. (This probably solves the issue of rollover funds paying for miner fees as well.) The main downside is...your change output is now locked for a day. 😬 So if you're a newly-funded renter with 1 huge output, and you want to form 50 contracts, you're still gonna need a setup transaction.

PR #47, opened 2022-02-10
types: Move signatures into FileContract type

lukechampine:
This brings `FileContract` into parity with other signed types, like `Attestation`. It will also make the upcoming contract renewal changes a bit cleaner.

Historically, the initial file contract hasn't been signed because it's redundant: each party is supplying inputs to the contract transaction, and they have to sign those inputs. But it bugs me that someone could broadcast a transaction that *claims* to form a contract with a particular host (by using their public key), even though the host actually knew nothing about it. Obviously, the host would then "fail" the storage proof, so these phony contracts could negatively impact a host's reputation. Requiring signatures on the initial contract prevents this.


lukechampine:
good catch -- updated

PR #48, opened 2022-02-11
consensus: Disallow contracts from being updated multiple times per block

lukechampine:
I'm not sure how multiple updates would affect the accumulator today, but it probably isn't pretty. (Meaning, it probably doesn't cause a validation error, but instead produces an invalid `ApplyUpdate` that can screw up your proofs.)

We *could* make an extra effort to allow this behavior; specifically, we would have to ensure that only the last revision/resolution is used for accumulator updates. But why not move that logic to the txpool instead? That is, if the txpool contains 2 transactions that revise the same contract, why not discard the one with the lower revision number? It would just waste block space anyway.

On that note, we should update `(ValidationContext).ValidateTransaction` to catch double-spends (and double-updates) within a single transaction, so that the txpool can reject them properly. And we should add the revision-selection code to the txpool, of course.

@n8maninger, can you write a test for this?

### DIFF
--- a/consensus/validation_test.go
+++ b/consensus/validation_test.go
@@ -3,6 +3,8 @@ package consensus
 import (
 	"encoding/binary"
 	"math"
+	"reflect"
+	"strings"
 	"testing"
 	"time"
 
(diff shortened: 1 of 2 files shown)


consensus/validation_test.go
```diff
@@ -1108,3 +1110,221 @@ func TestValidateBlock(t *testing.T) {
...
+		for _, set := range revisionTransactionTests {
+			if err := vc.ValidateBlock(mineBlock(vc, b, set...)); err == nil {
+				t.Fatal("expected invalid block error")
+			} else if !strings.Contains(err.Error(), "multiple times (previously updated in transaction") { // TODO: use errors.Is?
+				t.Fatal("expected multiple update error")
+			}
+		}
+	}
+
+	// test that a block will not be accepted if a revision is signed incorrectly
```
lukechampine, line 1245: `TestValidateTransaction` and `TestValidateTransactionSet` should already cover these sorts of cases; for this PR I only really care about testing a block with multiple revisions and/or resolutions of the same contract

PR #49, opened 2022-02-11
types: Make 'early close' an explicit form of contract resolution

lukechampine:
Previously, a `FileContractRevision` could actually be a `FileContractResolution` if it satisfied certain properties. That felt a little too magic to me, so instead I moved the "early close" case into the `FileContractResolution` type. Now, a revision always revises, and a resolution always resolves.

...well, except that a resolution can also revise. But still, I think this is an improvement.

### DIFF
--- a/net/rhp/contracts.go
+++ b/net/rhp/contracts.go
@@ -63,18 +63,6 @@ func PaymentRevision(fc types.FileContract, amount types.Currency) (types.FileCo
 	return fc, nil
 }
 
-// ClearingRevision returns a new file contract with the revision
-// number set to uint64 max, the file fields set to their zero value, and the
-// missed proof outputs set to the valid proof outputs.
-func ClearingRevision(fc types.FileContract) types.FileContract {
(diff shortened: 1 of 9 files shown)


net/rhp/contracts.go
```diff
@@ -63,18 +63,6 @@ func PaymentRevision(fc types.FileContract, amount types.Currency) (types.FileCo
 	return fc, nil
 }
 
-// ClearingRevision returns a new file contract with the revision
-// number set to uint64 max, the file fields set to their zero value, and the
-// missed proof outputs set to the valid proof outputs.
-func ClearingRevision(fc types.FileContract) types.FileContract {
```
lukechampine, line 69: we could keep this around as:
```go
func FinalizationRevision(fc types.FileContract) types.FileContract {
    fc.RevisionNumber = types.MaxRevisionNumber
    return fc
}
```
but it's so trivial that I think we should just inline it where needed

PR #51, opened 2022-02-14
First-class contract renewal

lukechampine:
This PR reifies contract renewals as a fourth type of contract resolution. The old contract is finalized (i.e. its revision number is maxed out) and marked resolved; the new contract is created; and importantly, a variable amount of SC may be "rolled over" from the old contract into the new contract.

After https://github.com/SiaFoundation/core/pull/43, I wasn't sure if rolling over funds like this could be done safely. But after talking to David about it, I feel pretty confident that this is okay. The key point is that *funds never leave a contract without a timelock*. You can roll funds from contract A->B (and B->C, and C->D), but when the final contract ends and the outputs are released, they'll be subject to the standard maturity delay. Likewise, the portion of the funds that *aren't* rolled over are also timelocked.

Rollover works like this: when you renew a contract, you can rollover any amount, up to the old contract's valid payout. So if the old contract resolved with the renter receiving 10 SC and the host receiving 90 SC, the renter can rollover up to 10 SC and the host can rollover up to 90 SC. The rollover amounts are (effectively) subtracted from the old contract's valid payout before it resolves. So if the renter rolls over 5 SC and the host rolls over 50 SC, then the renewal will create a 5 SC output owned by the renter, and a 40 SC output owned by the host (both timelocked). The other 55 SC will contribute towards the cost of the new contract.

There are two possibilities to consider here: either the new contract can costs more than the old contract, or it costs less (or exactly the same). If it costs more, then additional funding is required; this must be supplied in the form of SiacoinInputs, just like a normal contract. If it's less than or equal, then the rollover can cover the full cost of the new contract. In both cases, the renter and host are free to choose exactly how much they want to roll over. Another thing worth mentioning is that the old contract's payouts will be created even if they're zero-valued (which is likely to occur if the new contract costs more). I felt that this was simpler, although it does increase the total number of outputs somewhat.

Final note: since file contracts now contain their own signatures, in theory you could extract the final (or initial) contract from a renewal and broadcast it by itself. I don't *think* this is a problem, but maybe I didn't think about it deeply enough. If it does become a problem, I guess we can just zero-out the signatures? This sort of thing makes me wonder if https://github.com/SiaFoundation/core/pull/47 was a bit hasty...

### DIFF
--- a/consensus/update.go
+++ b/consensus/update.go
@@ -115,8 +115,12 @@ func updatedInBlock(vc ValidationContext, b types.Block, apply bool) (scos []typ
 		}
 		for _, fcr := range txn.FileContractResolutions {
 			fce := fcr.Parent
-			if apply && fcr.HasFinalization() {
-				fce.FileContract = fcr.Finalization
+			if apply {
+				if fcr.HasRenewal() {
+					fce.FileContract = fcr.Renewal.FinalRevision
+				} else if fcr.HasFinalization() {
+					fce.FileContract = fcr.Finalization
+				}
 			}
 			resolved = append(resolved, fce)
 			addLeaf(merkle.FileContractLeaf(fce, apply))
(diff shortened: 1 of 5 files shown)


lukechampine:
Final final note: maybe renewals should be their own `Transaction` field? Not sure if lumping them in with the other resolution types is the best approach.

PR #52, opened 2022-02-15
Cover Attestations in SigHash

n8maninger:
Changes `ValidationContext.SigHash` to use `ContractSigHash` instead of encoding the `FileContract` to prevent contract signatures from invalidating the transaction signatures and adds the `Attestations` field.

### DIFF
--- a/consensus/validation.go
+++ b/consensus/validation.go
@@ -265,6 +265,10 @@ func (vc *ValidationContext) SigHash(txn types.Transaction) types.Hash256 {
 	for _, fcr := range txn.FileContractResolutions {
 		fcr.Parent.ID.EncodeTo(h.E)
 		fcr.StorageProof.WindowStart.EncodeTo(h.E)
+		fcr.Finalization.EncodeTo(h.E)
+	}
+	for _, a := range txn.Attestations {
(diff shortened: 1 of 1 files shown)


lukechampine:
Is this necessary because of how signatures are exchanged in the RHP?

I guess, in general, signatures rarely need to cover other signatures. Transaction signatures don't cover other transaction signatures, for example.

lukechampine:
Perhaps we could simplify the logic by zeroing out the signatures instead of skipping them. Then we wouldn't need separate functions. Just need to be careful not to modify the *actual* signature...

n8maninger:
Yes, we could work around this by exchanging the `FileContract` signatures earlier.

1. Renter sends signed `FileContract` as part of the initial `RPCContractRequest`
2. Host adds inputs to the transaction, signs `FileContract`, and sends both to the renter.
3. Renter signs inputs and sends signatures
4. Host adds input signatures, broadcasts, and sends to renter

lukechampine:
That sounds preferable to me. The current RHP only sends them together because Sia v1 doesn't have explicit contract signatures, only generic transaction signatures.

PR #53, opened 2022-02-15
Separate contract signature validation

n8maninger:
Removes signature validation from the validation functions, signatures would need to be validated externally. The primary motivation was changing `ValidateContractFinalization` to validate all fields except the signature fields to short-circuit the renew RPC before the host has the renter signatures, like `ValidateRenewal` and `ValidateContractFormation`. However, I'm unsure if the change makes sense for the other validation methods, like `ValidatePaymentRevision`, which would have the signatures immediately.

### DIFF
--- a/net/rhp/contracts.go
+++ b/net/rhp/contracts.go
@@ -3,7 +3,6 @@ package rhp
 import (
 	"errors"
 	"fmt"
-	"math"
 
 	"go.sia.tech/core/consensus"
 	"go.sia.tech/core/types"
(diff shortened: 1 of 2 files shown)


net/rhp/contracts.go
```diff
@@ -139,100 +138,120 @@ func ValidateContractRenewal(existing, renewal types.FileContract, currentHeight
 	return nil
 }
 
-func validateStdRevision(vc consensus.ValidationContext, current, revision Contract) error {
+// ValidateContractFinalization verifies that the revision locks the current
+// contract by setting its revision number to the maximum legal value, no other
+// fields should change. Signatures are not validated.
+func ValidateContractFinalization(current, final types.FileContract) error {
```
lukechampine, line 144: could this be implemented like so?
```go
func ValidateContractFinalization(current, final types.FileContract) error {
    exp := current
    exp.RevisionNumber = types.MaxRevisionNumber
    if exp != final {
        return errors.New("finalized contract changes fields other than the revision number")
    }
    return nil
}
```

n8maninger: The signatures need to be zeroed out and a generic "something changed" error is a little less friendly, but it would work

lukechampine: true, it's a bit less friendly. On the other hand, it's future-proof against changes to the `FileContract` type.

idk, it's probably fine either way.

PR #55, opened 2022-02-16
Send contract signatures earlier

n8maninger:
Changes the RHP types to send the host's contract signatures before the Siacoin input signatures as described in #52


lukechampine:
I think we'll want to make even more aggressive changes to these RPCs, like requiring a single txn (instead of a slice), disallowing parents, disallowing multiple outputs, etc. This looks fine for now, though.

PR #57, opened 2022-02-20
Refine FileContract output fields

lukechampine:
This implements the change proposed in https://github.com/SiaFoundation/siad/pull/83#discussion_r810456991. I opted for three fields instead of four; `HostOutput` now includes `CollateralAvailable`, with `CollateralRisked` renamed to simply `Collateral`.

Thoughts? I think this might need another small tweak or two, but overall it seems to simplify the output handling quite a bit (particularly sanity checking the missed outputs -- a number of invalid states are now impossible to construct).


net/rhp/contracts.go
```diff
@@ -47,29 +47,27 @@ func (c *Contract) MaxLen() uint64 {
...
-// amount subtracted from the missed host output.
+// amount subtracted from the host output. The revision number is incremented.
 func FinalizeProgramRevision(fc types.FileContract, burn types.Currency) (types.FileContract, error) {
-	if fc.MissedHostOutput.Value.Cmp(burn) < 0 {
+	if fc.HostOutput.Value.Cmp(burn) < 0 {
 		return fc, errors.New("not enough funds")
 	}
 	fc.RevisionNumber++
-	fc.MissedHostOutput.Value = fc.MissedHostOutput.Value.Sub(burn)
+	fc.HostOutput.Value = fc.HostOutput.Value.Sub(burn)
```
lukechampine, line 70: this is the only part of the diff that I felt uneasy about. Is it correct to simply destroy value here? From the comment on https://github.com/SiaFoundation/siad/blob/master/types/filecontracts.go#L206, it seems like this revision is only supposed to be used in tandem with EAs somehow. Need more clarity here.

net/rhp/contracts.go
```diff
@@ -240,18 +231,12 @@ func ValidatePaymentRevision(current, revision types.FileContract, amount types.
...
-	case revision.ValidHostOutput.Value.Sub(amount) != current.ValidHostOutput.Value:
-		return errors.New("host valid output value should increase by the amount")
-	case revision.MissedRenterOutput.Value.Add(amount) != current.MissedRenterOutput.Value:
-		return errors.New("renter missed output value should decrease by the amount")
-	case revision.ValidRenterOutput.Value.Add(amount) != current.ValidRenterOutput.Value:
-		return errors.New("renter valid output value should decrease by the amount")
+	case revision.RenterOutput.Value.Add(amount) != current.RenterOutput.Value:
+		return errors.New("renter output value should decrease by the amount")
+	case revision.MissedHostValue != current.MissedHostValue.Add(amount):
+		return errors.New("host missed output value should decrease by the amount")
```
n8maninger, line 237: `should increase`

net/rhp/contracts.go
```diff
@@ -117,29 +112,29 @@ func ValidateContractRenewal(existing, renewal types.FileContract, currentHeight
...
 		return errors.New("revision number must be zero")
 	case renewal.Filesize != existing.Filesize:
-		return errors.New("file size must not change")
+		return errors.New("filesize must not change")
 	case renewal.FileMerkleRoot != existing.FileMerkleRoot:
-		return errors.New("file merkle root must not change")
+		return errors.New("file Merkle root must not change")
 	case renewal.WindowEnd < existing.WindowEnd:
-		return errors.New("renewal window end must be after existing window end")
+		return errors.New("renewal window end must be after current window end")
```
n8maninger, line 119: If we want to allow funding only renewals should this be `<=`?

net/rhp/contracts.go
```diff
@@ -117,29 +112,29 @@ func ValidateContractRenewal(existing, renewal types.FileContract, currentHeight
...
 	case renewal.WindowEnd < renewal.WindowStart+settings.WindowSize:
 		return errors.New("proof window is too small")
-	case renewal.ValidHostOutput.Address != settings.Address:
-		return errors.New("wrong address for host valid output")
-	case renewal.MissedHostOutput.Address != settings.Address:
-		return errors.New("wrong address for host missed output")
-	case renewal.ValidHostOutput.Value.Cmp(settings.ContractFee) < 0:
+	case renewal.HostOutput.Address != settings.Address:
+		return errors.New("wrong address for host output")
+	case renewal.HostOutput.Value.Cmp(settings.ContractFee) < 0:
```
n8maninger, line 128: Should this check `TotalCollateral` too? 🤔 
```go
renewal.HostOutput.Value.Cmp(settings.ContractFee.Add(renewal.TotalCollateral)) < 0
```

net/rhp/contracts.go
lukechampine: I think it's correct as written -- `renewal.WindowEnd == existing.WindowEnd` will *not* raise an error. The error message is a bit misleading, though

net/rhp/contracts.go
lukechampine: yeah I wasn't sure about this one, because renewal math. I *think* what you wrote it correct, though. Or should it be exactly equal, as in `ValidateContractFormation`?

n8maninger: It won't be exactly equal because we don't know the additional storage cost

PR #58, opened 2022-03-17
Add storage proof functions

n8maninger:
Adds the `us` functions to build storage proofs

### DIFF
--- a/net/rhp/merkle.go
+++ b/net/rhp/merkle.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"errors"
 	"io"
+	"math"
 	"math/bits"
 	"unsafe"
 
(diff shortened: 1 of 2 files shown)


net/rhp/merkle.go
```diff
@@ -251,6 +256,107 @@ func nextSubtreeSize(start, end int) int {
...
+		}
+	}
+	buildRange(0, start)
+	buildRange(end, math.MaxInt32)
+	return proof
+}
+
+// ConvertProofOrdering converts "left-to-right" proofs into the "leaf-to-root"
+// ordering used in consensus storage proofs.
+func ConvertProofOrdering(proof []types.Hash256, index int) []types.Hash256 {
```
lukechampine, line 342: I don't think we need this anymore. Sia v1 validates storage proofs in "leaf-to-root" order, whereas Sia v2 validates them "left-to-right" (matching how all other proofs are constructed in the RHP).

PR #59, opened 2022-03-17
Allow empty storage proofs within the proof window

n8maninger:
Hosts cannot currently retrieve funds for empty contracts until after the proof window unless the renter finalizes the contract. This PR allows contract resolutions with empty storage proofs when the contract has a zero Merkle root.

### DIFF
--- a/consensus/update.go
+++ b/consensus/update.go
@@ -216,6 +216,8 @@ func createdInBlock(vc ValidationContext, b types.Block) (sces []types.SiacoinEl
 				renter, host = fce.RenterOutput, fce.HostOutput
 			} else if fcr.HasFinalization() {
 				renter, host = fcr.Finalization.RenterOutput, fcr.Finalization.HostOutput
+			} else if fce.Filesize == 0 {
+				renter, host = fce.RenterOutput, fce.HostOutput
 			} else {
 				renter, host = fce.RenterOutput, fce.MissedHostOutput()
 			}
--- a/consensus/update_test.go
+++ b/consensus/update_test.go
@@ -541,6 +541,7 @@ func TestFileContracts(t *testing.T) {
 		merkle.StorageProofLeafHash(data[64:]),
 	)
 	finalRev.Revision.RevisionNumber++
+	finalRev.Revision.Filesize = uint64(len(data))
 	contractHash = sau.Context.ContractSigHash(finalRev.Revision)
 	finalRev.Revision.RenterSignature = renterPrivkey.SignHash(contractHash)
 	finalRev.Revision.HostSignature = hostPrivkey.SignHash(contractHash)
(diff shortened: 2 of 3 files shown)


lukechampine:
Might it makes more sense to treat this as a *missed* claim? If the contract is empty, the valid and missed outputs should be the same. Plus, we wouldn't need to include irrelevant data like the `WindowProof`. Then we could write:

```go
	leafIndex := vc.StorageProofLeafIndex(fc.Filesize, fcr.StorageProof.WindowStart, fcr.Parent.ID)
	if merkle.StorageProofRoot(fcr.StorageProof, leafIndex) != fc.FileMerkleRoot {
		return fmt.Errorf("storage proof %v has root that does not match contract Merkle root", i)
	}
} else {
	// missed outputs for empty contracts can be claimed at any point after WindowStart;
	// non-empty contracts can only be claimed after WindowEnd
	if fc.Filesize == 0 && vc.Index.Height < fc.WindowStart {
		return fmt.Errorf("file contract resolution %v attempts to claim missed outputs for empty contract, but proof window (%v - %v) has not begun", i, fc.WindowStart, fc.WindowEnd)
	} else if vc.Index.Height <= fc.WindowEnd {
		return fmt.Errorf("file contract resolution %v attempts to claim missed outputs, but proof window (%v - %v) has not expired", i, fc.WindowStart, fc.WindowEnd)
	}
}
```

n8maninger:
That does look better. The logic would need to be reversed though:
```go
if vc.Index.Height < fc.WindowStart {
	return fmt.Errorf("file contract resolution %v attempts to claim missed outputs, but proof window (%v - %v) has not begun", i, fc.WindowStart, fc.WindowEnd)
} else if fc.Filesize != 0 && vc.Index.Height <= fc.WindowEnd {
	return fmt.Errorf("file contract resolution %v attempts to claim missed outputs for non-empty contract, but proof window (%v - %v) has not expired", i, fc.WindowStart, fc.WindowEnd)
}
```

n8maninger:
@lukechampine, is it guaranteed that the missed proof outputs will be equal on an empty contract, though? If all sectors are dropped, the `Filesize` and `FileMerkleRoot` fields will reset, but the storage revenue is still in the valid output.

lukechampine:
hmm, true. Perhaps the original approach was best after all. I imagine this sort of resolution will be pretty uncommon anyway.

lukechampine:
Alternatively, we could modify `createdInBlock` as well, to create valid outputs instead of missed ones in this case. 🤔

n8maninger:
I think either way is fine, but not requiring the storage proof fields for an empty resolution makes more logical sense.

lukechampine:
ok, implemented. This approach also means you can get the valid outputs even after the proof window expires, which is nice.

n8maninger:
Looks good 👍🏻

PR #61, opened 2022-03-19
Fix ElementID Unmarshaling

chris124567:
I was working on https://github.com/SiaFoundation/siad/pull/86 and I kept getting this "wrong number of separators" error.  Eventually I realized it was coming from here.  Also, looking at some of the other types, shouldn't all of the UnmarshalTexts operate on pointer receivers?


lukechampine:
oof. Not sure how that made it past quality control. I'll push a fix.

The `ElementID` string representation isn't great; it looks too similar to a `ChainIndex`. It also doesn't tell you whether the source hash is a block or a transaction. We might want to revisit that before release.

PR #63, opened 2022-03-25
Host updates

n8maninger:
Updates the host interfaces with changes from siad-v2, adds read and write byte counters to `mux`, and additional spending info to the budget.

### DIFF
--- a/host/host.go
+++ b/host/host.go
@@ -68,7 +68,56 @@ type (
 		Cap() uint64
 	}
 
-	// A ContractStore stores contracts, metadata, and proofs for the host.
+	// A SettingsReporter returns the host's current settings.
+	SettingsReporter interface {
+		Settings() rhp.HostSettings
+	}
+
+	// A TransactionPool broadcasts transaction sets to miners for inclusion in
+	// an upcoming block.
+	TransactionPool interface {
+		AddTransaction(txn types.Transaction) error
+		RecommendedFee() types.Currency
+		Transactions() []types.Transaction
+	}
+
+	// A Wallet provides addresses and funds and signs transactions.
+	Wallet interface {
+		Address() types.Address
+		FundTransaction(txn *types.Transaction, amount types.Currency, pool []types.Transaction) ([]types.ElementID, func(), error)
+		SignTransaction(vc consensus.ValidationContext, txn *types.Transaction, toSign []types.ElementID) error
+	}
+)
+
+type (
+	// ContractState is the current lifecycle stage of a contract.
+	ContractState string
+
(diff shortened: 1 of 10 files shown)


host/host.go
```diff
@@ -68,24 +68,79 @@ type (
...
+
+type (
+	// ContractState is the current lifecycle stage of a contract.
+	ContractState string
+
+	// A Contract contains metadata on the current lifecycle stage of a file
+	// contract.
+	Contract struct {
+		rhp.Contract
+		Parent types.FileContractElement
```
lukechampine, line 100: perhaps these fields could be replaced with `Revision types.FileContractRevision`? Unless there's value in storing it as an `rhp.Contract`.

host/host.go
```diff
@@ -68,24 +68,79 @@ type (
...
 		Revise(rhp.Contract) error
 
 		// Roots returns the roots of all sectors stored by the contract.
 		Roots(types.ElementID) ([]types.Hash256, error)
 		// SetRoots sets the stored roots of the contract.
 		SetRoots(types.ElementID, []types.Hash256) error
+
+		// ContractAction calls contractFn on every contract in the store that
+		// needs a lifecycle action performed.
+		ContractAction(vc consensus.ValidationContext, contractFn func(consensus.ValidationContext, Contract)) error
```
lukechampine, line 143: seems weird to pass a `vc` which is then immediately passed back into the `contractFn` you supplied. It looks like `vc` is used to decide which contracts are actionable; maybe instead, we could do:
```go
ForEachContract(fn func(Contract)) error
```
and have `fn` be responsible for skipping over contracts that don't require any action.

net/mux/mux.go
```diff
@@ -364,6 +364,8 @@ type Stream struct {
 	established bool      // has the first frame been sent?
 	err         error
 	readBuf     []byte
+	readBytes   uint64
+	writeBytes  uint64
```
lukechampine, line 368: I don't think this is the right place to be tracking bandwidth metrics. Easy to wrap the conn, like `statsConn` in `us`. Usually you'll want to track other metrics too anyway, and wrapping allows you to gather stats on any type of connection.

host/host.go
n8maninger: Having the manager loop through each contract in the store feels bad from a scalability standpoint. Maybe it's best to use the index from the store instead of passing it in?

lukechampine: true. Maybe we can get away with `ContractAction(height uint64, fn func(Contract)) error`? Then make `fn` a closure around the current `vc`.

PR #66, opened 2022-04-01
ci: add go 1.18

n8maninger:
Adds Go 1.18 to the CI test versions. The `go.mod` is still 1.16, do we want to drop it?


lukechampine:
Little reason to bump the `go.mod` version until we need language features introduced in later versions. (There are some [other effects too](https://go.dev/doc/modules/gomod-ref#go), but AFAICT they aren't consequential.)

PR #67, opened 2022-04-06
Add test to ensure mux.Write blocks after large amounts of data are sent

chris124567:
This PR adds `TestSlowRead` to `mux_test.go` which ensures that the mux writer does not allow messages to simply pile up and will block after a certain point if they are not being accepted by the receiver.  This kind of scenario is obviously unlikely to show up in local testing instances but in practice connections can be unreliable so it is worth testing.

### DIFF
--- a/net/mux/mux_test.go
+++ b/net/mux/mux_test.go
@@ -115,6 +115,7 @@ func TestSlowRead(t *testing.T) {
 				return err
 			}
 			defer s.Close()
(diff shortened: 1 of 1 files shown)


lukechampine:
Revisiting this, I don't think there's a ton of value in this test. We can easily confirm by inspection that `Write` calls block instead of allocating more memory. My bad. :/

PR #69, opened 2022-04-30
Add SpendPolicy JSON unmarshaling

chris124567:
Adds support for parsing SpendPolicys encoded with SpendPolicy.String() and accordingly adds a JSON unmarshaler function.  Also adds a new `TestPolicyJSON` that tests an encoding "round trip" (marshaling then unmarshaling) for a bunch of policies.

### DIFF
--- a/types/policy.go
+++ b/types/policy.go
@@ -3,6 +3,9 @@ package types
 import (
 	"encoding/binary"
 	"encoding/hex"
+	"encoding/json"
+	"errors"
+	"fmt"
 	"math/bits"
 	"strconv"
 	"strings"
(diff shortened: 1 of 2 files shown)


lukechampine:
Nice. This is good enough to unblock Alex, so I'll merge it now. But I'd like to see you take a stab at a more robust parser. Traditionally, that would look something like this (simplified):

```go
func ParseSpendPolicy(s string) (SpendPolicy, error) {
    sp, rem, err := parseSpendPolicy(s)
    if err != nil {
        return SpendPolicy{}, err
    } else if rem != "" {
        return SpendPolicy{}, fmt.Errorf("trailing bytes: %q", rem)
    }
    return sp, nil
}

func parseSpendPolicy(s string) (SpendPolicy, string, error) {
    nextToken := func() string {
        s = strings.TrimSpace(s)
        i := strings.IndexAny(s, "(),[]")
        t := s[:i]
        s = s[i:]
        return t
    }
    consume := func(b byte) {
        s = strings.TrimSpace(s)
        if s[0] != b {
            err = fmt.Errorf("expected %c, got %c", b, s[0])
        }
        s = s[1:]
    }

    typ := nextToken()
    expect('(')
    switch typ {
    case "above":
        n, err := strconv.ParseUint(nextToken(), 10, 64)
        sp = PolicyAbove(n)
    case "pk":
        // etc.
    }
    expect(')')
}
```

PR #70, opened 2022-04-30
v2: use a more robust parser in ParseSpendPolicy

chris124567:
Replaces the rather complicated string indexing of the old ParseSpendPolicy with a more standard recursive parser.

### DIFF
--- a/types/policy.go
+++ b/types/policy.go
@@ -1,11 +1,11 @@
 package types
 
 import (
+	"bytes"
 	"encoding/binary"
 	"encoding/hex"
-	"encoding/json"
-	"errors"
 	"fmt"
+	"io"
 	"math/bits"
 	"strconv"
 	"strings"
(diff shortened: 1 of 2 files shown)


types/policy.go
```diff
@@ -170,151 +170,136 @@ func (p SpendPolicy) String() string {
...
+		s = strings.TrimSpace(s)
+		i := strings.IndexAny(s, "(),[]")
+		if i == -1 {
+			return ""
+		}
+		t := s[:i]
+		s = s[i:]
+		return t
 	}
+	consume := func(b byte) error {
```
lukechampine, line 195: consider using a "sticky error" here so that you don't have to check the return value of `consume` at each callsite

lukechampine, line 195: consider using a "sticky error" here so that you don't have to check the return value of `consume` at each callsite

types/policy.go
```diff
@@ -170,151 +170,136 @@ func (p SpendPolicy) String() string {
...
-		}
 
-		return PolicyThreshold(uint8(n), policies), nil
-	case "uc":
-		comma := strings.IndexByte(between, ',')
-		if comma == -1 {
-			return SpendPolicy{}, errors.New("unlock conditions policy missing comma")
+			// last policy in list will not have comma after it
+			// so don't check for error
+			consume(',')
```
lukechampine, line 252: kinda janky... this will accept `thresh(1,above(2),)` as valid, won't it?

the standard solution to this problem is to add a `peek` function. Then you can write:
```go
for len(s) > 0 && peek() != ']' {
    // ... parseSpendPolicy ...
    
    if peek() != ',' {
        break
    }
    consume(',')
}
consume(']')
```

lukechampine, line 252: kinda janky... this will accept `thresh(1,above(2),)` as valid, won't it?

the standard solution to this problem is to add a `peek` function. Then you can write:
```go
for len(s) > 0 && peek() != ']' {
    // ... parseSpendPolicy ...
    
    if peek() != ',' {
        break
    }
    consume(',')
}
consume(']')
```

lukechampine:
In addition to the "roundtrip" JSON test you added (which takes a `SpendPolicy`, marshals it, unmarshals it, and compares the two), we should also add tests that just marshal (and compare to a string) and just unmarshal (from a known-good string)

lukechampine:
In addition to the "roundtrip" JSON test you added (which takes a `SpendPolicy`, marshals it, unmarshals it, and compares the two), we should also add tests that just marshal (and compare to a string) and just unmarshal (from a known-good string)

chris124567:
Should be good to go.

lukechampine:
tweaked it a bit further. Quite happy with how this turned out 👍🏻

PR #74, opened 2022-05-05
v2: Add encoding tests in rhp

chris124567:
Adds encoding roundtrip style test that brings coverage in rpc.go up to 72% and settings.go to 97%.


net/rhp/rpc_test.go
```diff
@@ -0,0 +1,88 @@
...
+		e := types.NewEncoder(&buf)
+		randVal, ok := quick.Value(reflect.TypeOf(val), rand.New(rand.NewSource(0)))
+		if ok {
+			randVal.Interface().(types.EncoderTo).EncodeTo(e)
+		} else {
+			val.EncodeTo(e)
+		}
+		e.Flush()
+
+		decval := reflect.ValueOf(val)
```
lukechampine, line 80: I worry about this aliasing pointers; to be safe, we should use `reflect.New(reflect.TypeOf(val))`

net/rhp/rpc_test.go
```diff
@@ -0,0 +1,88 @@
...
+	}
+
+	for _, val := range tests {
+		var buf bytes.Buffer
+		e := types.NewEncoder(&buf)
+		randVal, ok := quick.Value(reflect.TypeOf(val), rand.New(rand.NewSource(0)))
+		if ok {
+			randVal.Interface().(types.EncoderTo).EncodeTo(e)
+		} else {
+			val.EncodeTo(e)
```
lukechampine, line 76: does this mean that if `quick.Value` can't generate a value, we just use a zero-ed value instead?

I'd prefer to `t.Error` here. We should be generating values for all types.

net/rhp/rpc_test.go
```diff
@@ -0,0 +1,88 @@
...
+	"bytes"
+	"math/rand"
+	"reflect"
+	"testing"
+	"testing/quick"
+
+	"go.sia.tech/core/types"
+)
+
+func (p *HostSettings) Generate(rand *rand.Rand, size int) reflect.Value {
```
lukechampine, line 13: `// Generate implements quick.Generator.`

also, you can use `func (*HostSettings)` when the receiver variable is unused.

lukechampine:
Looks like this caught some encoding bugs, nice 👌

I'm gonna start cracking down on commit hygiene -- pls squash this into one commit, and prefix the commit message with `net/rhp`

chris124567:
Should be good to go.

PR #75, opened 2022-05-07
net/rhp: add contracts tests

chris124567:
Adds tests that try to trigger all of the error states in ValidateContractSignatures, ValidateContractFormation, ValidateContractRenewal, ValidateContractFinalization, ValidateProgramRevision, ValidatePaymentRevision.  Brings contracts.go coverage to over 90%.

### DIFF
--- a/net/rhp/contracts_test.go
+++ b/net/rhp/contracts_test.go
@@ -140,88 +140,79 @@ func TestValidateContractRenewalFinalization(t *testing.T) {
 	}
 
 	formChanges := []struct {
-		change func(fc types.FileContract) types.FileContract
-		error  string
+		corrupt func(fc *types.FileContract)
+		desc    string
 	}{
 		{
-			func(fc types.FileContract) types.FileContract {
+			func(fc *types.FileContract) {
 				fc.Filesize = 5
-				return fc
 			},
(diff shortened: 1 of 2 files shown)


net/rhp/contracts_test.go
```diff
@@ -0,0 +1,603 @@
...
+		t.Fatal(err)
+	}
+
+	renewalChanges := []struct {
+		change func(existing, renewal types.FileContract) types.FileContract
+		error  string
+	}{
+		{
+			func(existing, renewal types.FileContract) types.FileContract {
+				renewal.HostPublicKey[0] ^= 255
```
lukechampine, line 244: careful -- if `types.FileContract` contained a slice, and you mutated that, then all subsequent tests would fail with the same error.

you *should* be fine here, but consider making a copy of `renewal` for each test case and then passing a pointer to the `change` function. You might also want to compare the exact error string; this is usually an anti-pattern, but frankly error strings don't change very often and this would ensure that we're hitting the errors we want to hit.

lastly, let's rename the struct fields to be consistent with https://github.com/SiaFoundation/core/blob/master/consensus/validation_test.go#L402

chris124567:
This should be good.

PR #80, opened 2023-02-03
rhp/v3: add missing instruction costs

n8maninger:
+ Adds missing RHPv3 instruction costs
+ Adds `ResourceCost` to be explicit about where the cost comes from.
+ Doesn't include memory or time costs from `siad`; need to discuss what values are appropriate or if we want them to be included.
+ Changes how bandwidth costs are calculated in the MDM to only include the inputs and outputs, not the protocol overhead.

### DIFF
--- a/rhp/v3/rhp.go
+++ b/rhp/v3/rhp.go
@@ -5,8 +5,17 @@ import (
 	"time"
 
 	"go.sia.tech/core/consensus"
+	rhpv2 "go.sia.tech/core/rhp/v2"
 	"go.sia.tech/core/types"
 	"lukechampine.com/frand"
+)
+
+const (
+	blocksPerYear     = 365 * 144
+	registryEntrySize = 256
+
+	// SectorSize is the size of one sector in bytes.
+	SectorSize = rhpv2.SectorSize
 )
 
 // An Account is a public key used to identify an ephemeral account on a host.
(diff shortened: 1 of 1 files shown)


rhp/v3/rhp.go
```diff
@@ -5,10 +5,19 @@ import (
...
 	"go.sia.tech/core/types"
 	"lukechampine.com/frand"
 )
 
+const (
+	blocksPerYear     = 365 * 144
+	registryEntrySize = 256
+
+	// SectorSize is the size of one sector in bytes.
+	SectorSize = rhpv2.SectorSize
```
lukechampine, line 18: I resisted doing this in other places. I think it's best to stick with a single definition, so that there's no mixing and matching.

I suppose there's an argument to made that each RHP should define its own constants. However, RHPv3 isn't *really* it's own protocol -- it's an extension of RHPv2. For example, RHPv3 code makes use of the `rhpv2.HostSettings` type, and `hostd` has a top-level `rhp` package that is used for both v2 and v3.

rhp/v3/rhp.go
```diff
@@ -87,134 +96,145 @@ func PayByContract(rev *types.FileContractRevision, amount types.Currency, refun
...
+		// host to execute the instruction. It is freed if the instruction
+		// fails.
+		Collateral types.Currency
+		// Egress is the cost to send the instruction's output to the renter. It
+		// is not refunded if the program fails.
+		Egress types.Currency
+		// Ingress is the cost to receive the instruction's input from the
+		// renter. It is not refunded if the program fails.
+		Ingress types.Currency
+	}
```
lukechampine, line 120: how about:

```go
ResourceUsage struct {
    Base types.Currency
    Collateral types.Currency
    Storage uint64
    Ingress uint64
    Egress uint64
}
```
Then, methods like `AppendSectorCost` can become standalone `AppendSectorUsage` functions, and we can have a single `func(pt HostPriceTable, ru ResourceUsage) types.Currency` that multiplies each field by its corresponding cost.

Regardless, I imagine we'll want an `Add` method for adding usages/costs together, right? For programs with multiple instructions?

lukechampine:
> Changes how bandwidth costs are calculated in the MDM to only include the inputs and outputs, not the protocol overhead.

Fully support this. I always thought including the protocol bytes was silly -- if you're gonna do that, you have to account for *every* protocol beneath you, all the way down to ethernet!

rhp/v3/rhp.go
n8maninger: I don't have a strong opinion either way, but you'll still have outliers like `ReadOffsetCost` and `ReadSectorCost` which require `length`

PR #83, opened 2023-02-13
Fix newCollateral math in ContractRenewalCollateral

ChrisSchinnerl:
This PR fixes the math in `ContractRenewalCollateral` and `ContractFormationCollateral`.

1. both now use the same math to compute collateral given the `renterFunds` which we are about to put into a contract.
2. `costPerByte` now accounts for the storage duration. This should drive the cost up significantly and reduce the amount of collateral we put into contracts.
3. The collateral capping math in `ContractRenewalCollateral` now subtracts the delta from `newCollateral` which fixes the case of hitting the `MaxCollateral` with a `baseCollateral` that is smaller than `newCollateral`. Previously the `newCollateral` returned in that case was far too small.
4. Computing the collateral now multiplies the collateral by the duration of the contract as well. Otherwise we only add enough collateral for a single block.

### DIFF
--- a/rhp/v2/contracts.go
+++ b/rhp/v2/contracts.go
@@ -13,12 +13,20 @@ func ContractFormationCost(fc types.FileContract, contractFee types.Currency) ty
 }
 
 // ContractFormationCollateral returns the amount of collateral we add when forming a contract.
-func ContractFormationCollateral(expectedStorage, period uint64, host HostSettings) types.Currency {
-	hostCollateral := host.Collateral.Mul64(expectedStorage).Mul64(period)
-	if hostCollateral.Cmp(host.MaxCollateral) > 0 {
-		hostCollateral = host.MaxCollateral
+func ContractFormationCollateral(period uint64, renterFunds types.Currency, host HostSettings) types.Currency {
+	// calculate cost per byte. We use period / 2 for the duration to account
+	// for the fact that data is uploaded throughout the lifetime of the
+	// contract.
+	costPerByte := host.UploadBandwidthPrice.Add(host.StoragePrice.Mul64(period / 2)).Add(host.DownloadBandwidthPrice)
+	if costPerByte.IsZero() {
+		return types.ZeroCurrency
 	}
-	return hostCollateral
+	// calculate the collateral
+	collateral := host.Collateral.Mul(renterFunds.Div(costPerByte))
+	if collateral.Cmp(host.MaxCollateral) > 0 {
+		return host.MaxCollateral
+	}
+	return collateral
 }
 
 // PrepareContractFormation constructs a contract formation transaction.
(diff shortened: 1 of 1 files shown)


rhp/v2/contracts.go
```diff
@@ -13,12 +13,18 @@ func ContractFormationCost(fc types.FileContract, contractFee types.Currency) ty
 }
 
 // ContractFormationCollateral returns the amount of collateral we add when forming a contract.
-func ContractFormationCollateral(expectedStorage, period uint64, host HostSettings) types.Currency {
-	hostCollateral := host.Collateral.Mul64(expectedStorage).Mul64(period)
-	if hostCollateral.Cmp(host.MaxCollateral) > 0 {
-		hostCollateral = host.MaxCollateral
+func ContractFormationCollateral(period uint64, renterFunds types.Currency, host HostSettings) types.Currency {
+	// calculate cost per byte.
+	costPerByte := host.UploadBandwidthPrice.Add(host.StoragePrice.Mul64(period)).Add(host.DownloadBandwidthPrice)
```
n8maninger, line 18: Why are upload and download bandwidth needed in this calc?

ChrisSchinnerl: TLDR; (Turns out this is longer than expected): If you set your allowance to match your expected storage we should actually put less collateral into contracts this way. Because we don't do `allowance / contracts` when we compute `renterFunds`. Instead we only put a fraction (5-10%) of the allowance into contracts to not lock up all our money immediately.
Using `expectedStorage` might force the host to lock up more collateral than the renter can even pay for with the money it locked away. Of course the autopilot can use `expectedStorage` when computing `renterFunds` to further limit the amount of money it puts into contracts.

---
It depends. So before this change, the amount of collateral we require the host to put it depended on `expectedStorage`. Which is `expectedUploadedData / numContracts`. So if we expect to upload 50 TiB and we got 50 contracts, that's 1 TiB of collateral for the host.

After this change, `renterFunds` determines the amount of collateral. `renterFunds` is between 5% and 10% of the allowance divided by `numContracts`. So if storage costs 100 SC per TiB and you set your allowance to 5000 SC for 50 TiB, around 5-10 SC should end up in the contract as `renterFunds` and the equivalent of storage we can upload and download once within the period.

So assuming that the user sets the allowance according to their expected storage, this approach should put in less collateral since we only put in a small fraction of the renter's funds to leave some funds for contract refreshing in case of uneven contract use.

Essentially the goal of this change was to make `ContractFormationCollateral` and `ContractRenewalCollateral` consistent. Either both should use `renterFunds` or `expectedStorage`. I'm leaning towards this PR because we can always have the autopilot take `expectedStorage` into account when computing the `renterFunds`.

peterjan (approved):

rhp/v2/contracts.go
peterjan: I'm not sure if I would've gone for passing in the `renterFunds` but instead would've probably passed in `expectedStorage` into both. If it felt weird to you while writing it though I'm sure passing in `renterFunds` makes more sense 👍

n8maninger: Maybe `renterd` could calculate the expected storage accounting only for the 5%-10% of allowance going into the contract rather than the total expected storage? I'd still lean towards trying to use `expectedStorage` for both since it's the cleanest calculation unit. However, if it felt weird, it's probably best to stick to `renterFunds`.

peterjan (approved):

PR #85, opened 2023-02-15
chain: Add package

lukechampine:
This implements a chain manager that can accept blocks, attach them to the best chain, compute the resulting state and effects, and apply those effects to a database.

### DIFF
--- a/consensus/state.go
+++ b/consensus/state.go
@@ -44,7 +44,7 @@ type Store interface {
 	SiacoinOutput(id types.SiacoinOutputID) (types.SiacoinOutput, bool)
 	SiafundOutput(id types.SiafundOutputID) (types.SiafundOutput, types.Currency, bool)
 	FileContract(id types.FileContractID) (types.FileContract, bool)
-	MaturedSiacoinOutputs(height uint64) []SiacoinOutputDiff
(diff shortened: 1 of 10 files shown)


consensus/state.go
```diff
@@ -32,14 +33,18 @@ const (
 var hasherPool = &sync.Pool{New: func() interface{} { return types.NewHasher() }}
 
 // A Store stores blocks, siacoin outputs, siafund outputs, and file contracts.
+//
+// Store methods do not return errors. If a store encounters an error, it should
+// save the error and thereafter return empty values from all methods. It is the
+// caller's responsibility to check this error, and, if non-nil, discard any
+// results computed.
 type Store interface {
```
n8maninger, line 41: Should these and the functions that use them, like `ApplyState`, return errors instead? The additional context may help when tracking down potential bugs.

lukechampine: whoops, merged before I refreshed and saw comments.

`Store` originally had an `Err` method for retrieving the stored error, but you don't actually need that to be part of the consensus interface, because consensus doesn't care about it. Instead, we handle that in the concrete type. That's what `WithConsensus` does -- it gives you an implementation of `consensus.Store` with a sticky error, and returns it at the end.

PR #87, opened 2023-02-18
Add source to delayed outputs

n8maninger:


lukechampine:
I can't decide if I prefer the "Source" approach or:
```go
type TransactionDiff struct {
    ...
    SiafundClaimOutputs  []DelayedSiacoinOutputDiff
    ValidContractOutputs []DelayedSiacoinOutputDiff
    ...
}

type BlockDiff struct {
    ...
    MissedContractOutputs []DelayedSiacoinOutputDiff
    MinerOutput           DelayedSiacoinOutputDiff
    FoundationOutput      *DelayedSiacoinOutputDiff // non-nil if this is a subsidy block
    ...
}
```
The latter seems more consistent with the current approach -- e.g. we have separate fields for `CreatedSiacoinOutputs` and `SpentSiacoinOutputs`, rather than having a `Spent` bool in `SiacoinOutputDiff`.

The main advantage of merging the different kinds into a single slice is that you can conveniently loop over all of them. But I'm not sure when this would actually be useful in practice.

Thoughts?

n8maninger:
I don't think splitting `ImmatureSiacoinOutputs` makes as much sense as it does with `Created` and `Spent` since there's already a similar split with creation in `ImmatureSiacoinOutputs ` and deletion in `MaturedSiacoinOutputs`. Ideally, we'd be able to identify the source when they reach maturity, so we would need to split `MaturedSiacoinOutputs` too.

lukechampine:
Ah, right, I forgot about the maturing end. If we split by field, then the `Store` has to be responsible for tracking each kind separately (or adding a "Source" tag to its own internal encoding...)

ok, I'm convinced!

PR #88, opened 2023-02-20
Add Consensus/Chain Manager Test

chris124567:
Adds test that creates new chain.Manager and in memory DB, adds some blocks, ensures the block diffs are correct and checks that the SC and SF outputs from the transactions in the blocks are in the DB.

### DIFF
--- a/chain/db_test.go
+++ b/chain/db_test.go
@@ -8,25 +8,9 @@ import (
 
 	"go.sia.tech/core/consensus"
 	"go.sia.tech/core/types"
+	"go.sia.tech/core/wallet"
 	"lukechampine.com/frand"
 )
-
-// StandardUnlockConditions returns the standard unlock conditions for a single
-// Ed25519 key.
-func StandardUnlockConditions(pub types.PublicKey) types.UnlockConditions {
-	return types.UnlockConditions{
-		PublicKeys: []types.UnlockKey{{
-			Algorithm: types.SpecifierEd25519,
-			Key:       pub[:],
-		}},
-		SignaturesRequired: 1,
-	}
-}
-
-// StandardAddress returns the standard address for an Ed25519 key.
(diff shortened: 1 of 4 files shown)


chain/db_test.go
```diff
@@ -0,0 +1,293 @@
...
+	return cm.AddBlocks([]types.Block{*b})
+}
+
+type diffSubscriber struct {
+	cau *consensus.BlockDiff
+	cru *consensus.BlockDiff
+}
+
+func (s *diffSubscriber) ProcessChainApplyUpdate(cau *ApplyUpdate, mayCommit bool) error {
+	if cau == nil {
```
lukechampine, line 72: `cau` and `cru` will never be nil -- they are passed by pointer for efficiency (although perhaps they shouldn't be...)

chain/db_test.go
```diff
@@ -0,0 +1,293 @@
...
+		s.cau = nil
+		s.cru = nil
+		return nil
+	}
+	s.cau = nil
+	s.cru = &cru.Diff
+	return nil
+}
+
+func checkDBOutputsFunction(txn types.Transaction) func(DBTx) error {
```
lukechampine, line 93: can drop the `Function`

chain/db_test.go
```diff
@@ -0,0 +1,293 @@
...
+	"time"
+
+	"go.sia.tech/core/consensus"
+	"go.sia.tech/core/types"
+	"lukechampine.com/frand"
+)
+
+// StandardUnlockConditions returns the standard unlock conditions for a single
+// Ed25519 key.
+func StandardUnlockConditions(pub types.PublicKey) types.UnlockConditions {
```
lukechampine, line 16: perhaps these should be added to `wallet`?

lukechampine: looks like your `wallet.go` wasn't committed 😅

PR #89, opened 2023-02-20
Add support for multiplication with overflow to Currency

ChrisSchinnerl:
My node crashed due to an overflow in a `Mul` call so I extended `core` to allow for handling that.
I also thought it might be useful to have `MaxCurrency`. e.g. if multiplying storage and collateral results in an overflow, we could use `MaxCurrency` as the intermediary result for scoring instead of aborting.

### DIFF
--- a/chain/db_test.go
+++ b/chain/db_test.go
@@ -1,0 +1,269 @@
+package chain
+
+import (
+	"fmt"
+	"reflect"
+	"testing"
+	"time"
+
+	"go.sia.tech/core/consensus"
+	"go.sia.tech/core/types"
+	"go.sia.tech/core/wallet"
+	"lukechampine.com/frand"
+)
+
+// AppendTransactionSignature appends a TransactionSignature to txn and signs it
+// with key.
+func AppendTransactionSignature(cs consensus.State, key types.PrivateKey, parentID types.Hash256, txn *types.Transaction) {
+	sig := types.TransactionSignature{
+		ParentID:       parentID,
+		CoveredFields:  types.CoveredFields{WholeTransaction: true},
+		PublicKeyIndex: 0,
+	}
+	sigHash := key.SignHash(cs.WholeSigHash(*txn, sig.ParentID, sig.PublicKeyIndex, sig.Timelock, sig.CoveredFields.Signatures))
+	sig.Signature = sigHash[:]
+
+	txn.Signatures = append(txn.Signatures, sig)
+}
+
+// FindBlockNonce finds a block nonce meeting current target.
+func FindBlockNonce(cs consensus.State, b *types.Block) {
+	// ensure nonce meets factor requirement
+	for b.Nonce%cs.NonceFactor() != 0 {
+		b.Nonce++
+	}
+	for b.ID().CmpWork(cs.ChildTarget) < 0 {
+		b.Nonce += cs.NonceFactor()
+	}
+}
+
+func addBlock(cm *Manager, b *types.Block) error {
+	var minerAddress types.Address
+	frand.Read(minerAddress[:])
+
+	cs := cm.TipState()
+	b.MinerPayouts = []types.SiacoinOutput{{Address: minerAddress, Value: cs.BlockReward()}}
+	FindBlockNonce(cs, b)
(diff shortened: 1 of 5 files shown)


lukechampine:
I was wondering when this would bite us. :/

My *general* feeling on this is that, wherever we need to do non-trivial calculations involving `types.Currency`, we should be converting to (and from) `*big.Int`. Overflowing checking helps, but there are still calculations that simply can't be performed with `types.Currency` alone (specifically, calculations that require an intermediate value that is negative, or larger than 2^128).

On the other hand, converting to `*big.Int` is annoying, and kinda defeats the purpose of having super-fast 128-bit math, which makes me sad.

So here's an alternative solution: we introduce an `iCurrency` type (`i` for `intermediate`). The API would look something like this:
```go

type iCurrency struct {
	i      *big.Int
	lo, hi uint64
}

func (c Currency) Intermediate() iCurrency {
	return iCurrency{lo: c.Lo, hi: c.Hi}
}

func (c iCurrency) Add(x Currency) iCurrency   { /* ... */ }
func (c iCurrency) AddI(x iCurrency) iCurrency { /* ... */ }
func (c iCurrency) Add64(x uint64) iCurrency   { /* ... */ }
func (c iCurrency) Sub(x Currency) iCurrency   { /* ... */ }
func (c iCurrency) SubI(x iCurrency) iCurrency { /* ... */ }
func (c iCurrency) Sub64(x uint64) iCurrency   { /* ... */ }
// etc.

func (c iCurrency) Result() (Currency, error) {
	if c.i == nil {
		return Currency{c.lo, c.hi}, nil
	} else if c.i.Sign() < 0 {
		return ZeroCurrency, errors.New("value cannot be negative")
	} else if c.i.BitLen() > 128 {
		return ZeroCurrency, errors.New("value overflows Currency representation")
	}
	return NewCurrency(c.i.Uint64(), new(big.Int).Rsh(c.i, 64).Uint64()), nil
}
```
`Currency` itself does not support any arithmetic -- you have to convert to `iCurrency` first. If a calculation involving an `iCurrency` would overflow or underflow, we convert to a `big.Int` and use` c.i` thereafter. Otherwise, we use the nice fast 128-bit math. When the calculation is done, you call `Result`.

One thing I haven't checked is how this affects performance. Specifically, does having a `*big.Int` pointer change anything wrt garbage collection? Hopefully not, but we should definitely measure it.

lukechampine:
Man, I've gone back and forth and this and there is just no perfect solution. :/

`iCurrency` is a pain to use -- having two separate types is just too clunky. But if we try to do everything inside *one* type, then there's no way to force users to call `Result` -- nothing stops you from calling `EncodeTo` on an intermediate value, unless we want to add a bunch of runtime panics.

I see two options:
1. Aggressively validate untrusted `Currency` values. Once you know you're working with "reasonable" values, arithmetic becomes safe again. There aren't many "boundaries" where we'd need to do this: we're already doing it in `consensus`, and AFAIK that just leaves the RHP code. However, there could be more such boundaries in the future -- including ones written by non-core-developers, who would have to be as careful as we are to prevent panics. And even core devs will likely miss a few checks, resulting in potential DoS vulnerabilities.
2. Give up on pure 128-bit `Currency` values, i.e. make `iCurrency` the default. This actually matches the protocol better, since *technically* `Currency` values sent over the wire can be much larger than 128 bits. The main downside is performance: all `Currency` values would have to be scanned by the GC, and there are a *lot* of `Currency` values in Sia. `*big.Int` can also cause DoS vulnerabilities if an attacker can cause you to create a really big one, though I'm not aware of any Sia code that would be susceptible to this. (Do we ever multiply two attacker-supplied `Currency` values together? That would be a prime target.)

I'm leaning towards 2, but there's one more point in favor of 1: all of these points apply to `uint64` values too, and what are we doing about that? In other words, we're already trusting ourselves to properly validate inputs in a lot of places. Adding `Currency` to that set doesn't change the equation much.

n8maninger:
It'd be sad to lose the performance of uint128 for the general case because one malicious host set their download price to `18TS/TB`. We'd need to add validation for values like that regardless.

peterjan (approved):

PR #94, opened 2023-02-25
RHP Encoding Fixes

peterjan:
This PR fixes some of the encoding issues in the new `rhp` code. I've only tested it with `ReadSector`. I decided to open a PR because I think Nate and I are working through the same issues so I figured we could consolidate our efforts into one PR/branch and get it merged. 

The main issues I think are:
- [ ] instructions should be length prefixed
- [ ] `writeObject` requires a hack-fix in the lenght prefix because MDM programs now include the data
- [ ] order of field encoding is off at times
- [ ] some fields are missing in the encoding

### DIFF
--- a/rhp/v3/encoding.go
+++ b/rhp/v3/encoding.go
@@ -217,6 +217,7 @@ func (r *FundAccountReceipt) EncodeTo(e *types.Encoder) {
 	r.Host.EncodeTo(e)
 	r.Account.EncodeTo(e)
 	r.Amount.EncodeTo(e)
+	e.WriteTime(r.Timestamp)
 }
 
 // DecodeFrom implements ProtocolObject.
--- a/rhp/v3/encoding.go
+++ b/rhp/v3/encoding.go
@@ -158,7 +158,6 @@ func (r *PayByContractRequest) EncodeTo(e *types.Encoder) {
 	}
 	r.RefundAccount.EncodeTo(e)
 	e.WriteBytes(r.Signature[:])
-	r.HostSignature.EncodeTo(e)
 }
 
 // DecodeFrom implements ProtocolObject.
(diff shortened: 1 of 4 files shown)


rhp/v3/transport.go
```diff
@@ -95,7 +95,16 @@ func (s *Stream) writeObject(resp *rpcResponse) error {
...
 	e.Flush()
 	b := buf.Bytes()
-	binary.LittleEndian.PutUint64(b[:8], uint64(len(b)-8))
+	length := len(b) - 8
+	// HACK: in siad, the ExecuteProgram RPC sends the ProgramData separately
+	// from the rest of the request object. This was a performance optimization
+	// relating to how siad handles encoding and length prefixes. Suffice it to
+	// say, that optimization is not necessary in core, but we still have to
+	// preserve compatibility with how siad does things.
+	if r, ok := resp.data.(*RPCExecuteProgramRequest); ok {
```
n8maninger, line 104: Should this be handled by exposing `Stream.Write` and removing the `Output` field? This feels really hacky

peterjan: That was also what I had in mind but I talked it over with Luke and he really does not want to expose `Stream.Write`, which I definitely understand. It's still writing an RPC object 🤷‍♂️ . I'm usually not really in favour of hacks like this but in this case I don't really mind, it's pretty clear cut.

lukechampine:
lgtm -- are we keeping this as a draft until it's been tested with hosts?

n8maninger:
> lgtm -- are we keeping this as a draft until it's been tested with hosts?

There are a couple more I’m working on from the host side.

rhp/v3/encoding.go
```diff
@@ -291,7 +291,6 @@ func (r *RPCExecuteProgramRequest) DecodeFrom(d *types.Decoder) {
 			d.SetErr(fmt.Errorf("unrecognized instruction id: %q", id))
 			return
 		}
-		_ = d.ReadPrefix() // unused length prefix
```
n8maninger: The argument length prefix is handled above in `instructionForID`

n8maninger:
This looks good to me. `hostd` successfully executes append and read sector programs from `siad` renters. Unless @peterjan has any more, LGTM

peterjan:
LGTM 👍 will leave the merging up to @lukechampine

PR #95, opened 2023-02-27
Add testnet tax override

n8maninger:
Temporarily adds testnet consts to `rhp/v2` for the tax hardfork. Enables testnet builds in hostd and renterd

### DIFF
--- a/rhp/v2/testnet.go
+++ b/rhp/v2/testnet.go
@@ -1,4 +1,4 @@
-//go:build !testing && !testnet
+//go:build testnet
 
 package rhp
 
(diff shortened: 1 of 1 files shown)


lukechampine:
👍🏻 

Really need a better long-term solution for this. Maybe something like:
```go
package consensus

type NetworkParams struct {
    BlockInterval time.Duration
    HardforkTax uint64
    HardforkStorageProof uint64
    // ... stuff from siad/types/constants.go, basically
}

var MainNet = NetworkParams{ /* ... */ }
var TestNet = NetworkParams{ /* ... */ }

type State struct {
    Network NetworkParams
    // ...
}
```

PR #96, opened 2023-02-28
Network parameters

lukechampine:
I took a stab at addressing the issue that I alluded to in https://github.com/SiaFoundation/core/pull/95. The approach I ended up taking was a bit different than what I originally suggested. The problem with embedding the entire `NetworkParams` struct within `State` is that now you have to encode/decode all of those parameters every time you encode/decode a `State`. This is egregiously wasteful; the consensus DB would end up storing hundreds of thousands of copies of the same parameters. I suppose we could have the DB override the encoding of `State`, but you'd also have to override the encoding of any type that *contains* `State` (e.g. `Checkpoint`), and this would affect all future users of `State`, not just us.

It occurred to me that we could store just an ID, rather than the full params, and use the ID to select from a set of parameters defined locally in `consensus`. Then I thought: why bother with the IDs at all, when we can already distinguish between mainnet and testnet using `(State).GenesisTimestamp`? So that's what this PR does: use the genesis timestamp to select the set of network parameters.

I don't *love* this approach; it feels "sneaky," in the same way that global variables are sneaky. In particular, I don't like defining mainnet vs. testnet parameters within `consensus` itself. But the alternative -- passing them in -- means that they have to be embedded in every `State` object (wasteful/cumbersome) or supplied as a separate argument (i.e. everywhere we pass `consensus.State`, we now *also* have to pass a `NetworkParams`... gross). I also don't see much practical use for passing a `NetworkParams` *other* than mainnet or testnet.

That reminds me: another thing to note here is that we've gone from "standard, dev, testing, testnet" to just "mainnet, testnet." I don't think we have any real use for `dev` and `testing` anymore. `dev` was for antfarm stuff (which we can/should do on testnet now). `testing` was mostly a way of lowering the difficulty so that we could mine long chains during automated testing -- but if you set a very low initial difficulty, you can mine a *lot* of blocks before the difficulty ramps up. `testing` also triggered hardforks much earlier, but that's true of `testnet` as well, and the new consensus design makes it trivial to "fast-forward" to any desired hardfork height anyway.

Mainnet and testnet share a lot of constants (e.g. `BlockInterval`), so I didn't bother moving all of them to `networkParams`. You could argue that they belong there anyway, but idk, I'd rather have `networkParams` only specify what actually differs between chains.

cc @n8maninger

### DIFF
--- a/consensus/state.go
+++ b/consensus/state.go
@@ -12,19 +12,47 @@ import (
 	"go.sia.tech/core/types"
 )
 
-const (
-	blocksPerYear = 144 * 365
+type networkParams struct {
+	hardforkHeightDevAddr      uint64
+	hardforkHeightTax          uint64
+	hardforkHeightStorageProof uint64
+	hardforkHeightOak          uint64
+	hardforkHeightOakFix       uint64
+	hardforkHeightASIC         uint64
+	hardforkHeightFoundation   uint64
 
-	hardforkDevAddr      = 10000
-	hardforkTax          = 21000
-	hardforkStorageProof = 100000
-	hardforkOak          = 135000
-	hardforkOakFix       = 139000
-	hardforkASIC         = 179000
-	hardforkFoundation   = 298000
+	hardforkASICTotalTarget types.BlockID
+	hardforkASICTotalTime   time.Duration
+	minimumCoinbase         uint32 // in SC
+}
 
-	foundationSubsidyFrequency = blocksPerYear / 12
-)
+var mainnetParams = &networkParams{
+	hardforkHeightDevAddr:      10000,
+	hardforkHeightTax:          21000,
+	hardforkHeightStorageProof: 100000,
+	hardforkHeightOak:          135000,
+	hardforkHeightOakFix:       139000,
+	hardforkHeightASIC:         179000,
+	hardforkHeightFoundation:   298000,
+
+	hardforkASICTotalTarget: types.BlockID{8: 32},
+	hardforkASICTotalTime:   120000 * time.Second,
+	minimumCoinbase:         30000,
+}
+
+var testnetParams = &networkParams{
+	hardforkHeightDevAddr:      10000,
+	hardforkHeightTax:          2,
+	hardforkHeightStorageProof: 100000,
+	hardforkHeightOak:          10,
+	hardforkHeightOakFix:       12,
+	hardforkHeightASIC:         20,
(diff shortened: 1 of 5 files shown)


consensus/state.go
```diff
@@ -12,19 +12,47 @@ import (
...
+
+	hardforkASICTotalTarget: types.BlockID{8: 32},
+	hardforkASICTotalTime:   120000 * time.Second,
+	minimumCoinbase:         30000,
+}
+
+var testnetParams = &networkParams{
+	hardforkHeightDevAddr:      10000,
+	hardforkHeightTax:          2,
+	hardforkHeightStorageProof: 100000,
```
lukechampine, line 46: `hardforkHeightDevAddr` and `hardforkHeightStorageProof` are currently hard-coded in `modules/consensus`. Ideally `siad` would define them properly in `types.go` so that we could specify testnet heights for them. Of course, that would require resetting the testnet, so maybe not worth it?

n8maninger: I think fixing those two consts and resetting the testnet is a good idea

n8maninger: Looks like the dev address hardfork should be disabled on the testnet
```go
if build.Release != "testnet" && sfo.UnlockHash == gsa[len(gsa)-1].UnlockHash && blockHeight(tx) > 10e3 {
    sfo.UnlockHash = devAddr
}
```

Since this resets the initial siafund allocation to a hardcoded address, I'm not sure what the best way to have it activate on the testnet is. We'd likely want to remove the hardcoded `devAddr` too.

lukechampine: I would argue that siad's behavior is faulty here; it's using `gsa[len(gsa)-1]` as shorthand for the (old) dev address, but really it should be hardcoding it (which is what `core` does). If we hardcode it to the mainnet addr, it won't affect testnet at all.

n8maninger: The testnet already overrides the proof hardfork height too. I added to the hardcoded if statement instead of defining a new variable 😬 
```go
if (build.Release == "standard" && height < 100e3) || (build.Release == "testnet" && height < 5) || (build.Release == "testing" && height < 10) {
    return validStorageProofs100e3(tx, t)
}
```

n8maninger: We shouldn't need to reset the testnet even if we fix the logic in `siad`.

n8maninger: siad changes are in SiaFoundation/siad#156

PR #97, opened 2023-03-03
consensus: Use build tag for testnet

lukechampine:
Flip-flopping on #96 :|

I don't like build tags in general, but I'll allow it for this. It's better than branching through `s.params()` everywhere.

For posterity, I also investigated a "fork flag" approach, which worked like this:
```go
type ForkFlags uint64

const (
	ForkTestnet ForkFlags = iota
	ForkDevAddr
	ForkTax
	// ...
)

type State struct {
	ForkFlags ForkFlags
	// ...
}

// used like so:
func (s State) NonceFactor() uint64 {
	if s.ForkFlags&(1<<ForkASIC) != 0 {
		return 1009
	}
	return 1
}

// set like so:

type Network interface {
	ForkFlags(s State, b types.Block) ForkFlags
}

func ApplyState(s State, store Store, b types.Block, network Network) State {
	// ...
	return State{
		ForkFlags: network.ForkFlags(s, b),
	}
```

Basically, you pass a `Network` to ApplyState, and it sets the fork flags on the resulting `State`. So on mainnet (and testnet), we would pass a `Network` that sets flags based on `s.Index.Height`. In our automated tests, though, we could set the flags directly, instead of fast-forwarding to a particular height like we do now. Neat! Another cute feature is that you can encode the flags as a string, with emoji representing each flag.

Unfortunately, practice turned out to be uglier than theory. This approach *could* work, but it would require many more flags than you would expect. For example, the Foundation subsidy is paid out every 4380 blocks, starting at the hardfork height. But in this design, `State` doesn't know what the hardfork height is. So instead, we have to define a separate `ForkFoundationSubsidy` flag, and make `Network` responsible for setting that flag at the appropriate heights. Oh, but you also need a `ForkFoundationInitialSubsidy`, which is set exactly once, on the hardfork activation block itself! Similarly, you need two or three distinct flags for the ASIC hardfork, because there are specific actions that must occur on the activation block *and* on the block immediately prior to activation.

Like I said... it *could* work, but it forces you to move what *feels like* consensus code (e.g. Foundation subsidies) outside of the `consensus` package, which makes me uncomfortable. I got about halfway through the implementation before it started feeling so messy that I just got sad and scrapped it all. If I were implementing a *new* blockchain, I might take this approach, but it just doesn't work very well for the blockchain we have. :/

### DIFF
--- a/consensus/network_mainnet.go
+++ b/consensus/network_mainnet.go
@@ -1,0 +1,24 @@
+//go:build !testnet
+
+package consensus
+
+import (
+	"time"
+
+	"go.sia.tech/core/types"
+)
+
+const (
+	hardforkHeightDevAddr      = 10000
+	hardforkHeightTax          = 21000
+	hardforkHeightStorageProof = 100000
+	hardforkHeightOak          = 135000
+	hardforkHeightOakFix       = 139000
+	hardforkHeightASIC         = 179000
+	hardforkHeightFoundation   = 298000
+
+	hardforkASICTotalTime = 120000 * time.Second
+	minimumCoinbase       = 30000
+)
+
+var hardforkASICTotalTarget = types.BlockID{8: 32}
--- a/consensus/network_testnet.go
+++ b/consensus/network_testnet.go
@@ -1,0 +1,24 @@
+//go:build testnet
+
+package consensus
+
+import (
+	"time"
+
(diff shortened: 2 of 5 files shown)


lukechampine:
closing in favor of #98

PR #105, opened 2023-04-04
Add helpers for rhpv3 contract renewals

ChrisSchinnerl:
Adds the same helpers we already have for rhpv2 but slightly updated for rhpv3.

NOTE: `taxAdjustedPayout` is a copy-paste and exactly the same for rhpv2. Feel free to suggest a better place for it or to push a commit that moves it.

### DIFF
--- a/rhp/v3/contracts.go
+++ b/rhp/v3/contracts.go
@@ -1,0 +1,118 @@
+package rhp
+
+import (
+	"math/bits"
+
+	"go.sia.tech/core/consensus"
+	"go.sia.tech/core/types"
+)
+
+// ContractRenewalCost returns the cost of renewing a contract for the renter.
+// In other words, this is the amount of money that the renter needs to fund the
+// contract txn with.
+func ContractRenewalCost(cs consensus.State, pt HostPriceTable, fc types.FileContract, minerFee, basePrice types.Currency) types.Currency {
+	return fc.ValidRenterPayout().Add(pt.ContractPrice).Add(minerFee).Add(basePrice).Add(cs.FileContractTax(fc))
+}
+
+// PrepareContractRenewal constructs a contract renewal transaction.
+func PrepareContractRenewal(currentRevision types.FileContractRevision, hostAddress, renterAddress types.Address, renterKey types.PrivateKey, renterPayout, newCollateral types.Currency, hostKey types.PublicKey, pt HostPriceTable, endHeight uint64) (types.FileContract, types.Currency) {
+	hostValidPayout, hostMissedPayout, voidMissedPayout, basePrice := CalculateHostPayouts(currentRevision.FileContract, newCollateral, pt, endHeight)
+
+	return types.FileContract{
+		Filesize:       currentRevision.Filesize,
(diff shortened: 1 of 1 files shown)


rhp/v3/contracts.go
```diff
@@ -0,0 +1,118 @@
...
+	// if the contract height did not increase both prices are zero
+	if contractEnd := uint64(endHeight + pt.WindowSize); contractEnd > fc.WindowEnd {
+		timeExtension := uint64(contractEnd - fc.WindowEnd)
+		basePrice = basePrice.Add(pt.WriteStoreCost.Mul64(fc.Filesize).Mul64(timeExtension))
+		baseCollateral = baseCollateral.Add(pt.CollateralCost.Mul64(fc.Filesize).Mul64(timeExtension))
+	}
+
+	// calculate payouts
+	hostValidPayout := pt.ContractPrice.Add(basePrice).Add(baseCollateral).Add(newCollateral)
+	voidMissedPayout := basePrice.Add(baseCollateral)
```
n8maninger, line 57: This would cause the host to burn `pt.RenewContractCost`. Is that the behavior in `siad`?

rhp/v3/contracts.go
```diff
@@ -0,0 +1,118 @@
...
+
+	"go.sia.tech/core/consensus"
+	"go.sia.tech/core/types"
+)
+
+// ContractRenewalCost returns the cost of renewing a contract for the renter.
+// In other words, this is the amount of money that the renter needs to fund the
+// contract txn with.
+func ContractRenewalCost(cs consensus.State, pt HostPriceTable, fc types.FileContract, minerFee, basePrice types.Currency) types.Currency {
+	return fc.ValidRenterPayout().Add(pt.ContractPrice).Add(minerFee).Add(basePrice).Add(cs.FileContractTax(fc))
```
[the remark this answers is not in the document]
ChrisSchinnerl: `pt.RenewContractCost` is part of `basePrice` right now.

rhp/v3/contracts.go
n8maninger: It looks like that is the behavior in `siad`.

rhp/v3/contracts.go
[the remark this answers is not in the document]
n8maninger: Nevermind, it's included in `basePrice`

rhp/v3/contracts.go
```diff
@@ -0,0 +1,118 @@
...
+		var r uint64
+		if c.Hi < v {
+			_, r = bits.Div64(c.Hi, c.Lo, v)
+		} else {
+			_, r = bits.Div64(0, c.Hi, v)
+			_, r = bits.Div64(r, c.Lo, v)
+		}
+		return types.NewCurrency64(r)
+	}
+	sfc := (consensus.State{}).SiafundCount()
```
lukechampine, line 111: yikes. I guess I should fix this (in `rhp/v2` as well). It obviously breaks if we want to e.g. use a different siafund count on a testnet.

PR #110, opened 2023-04-18
RenterOutputIndex and HostOutputIndex Constants

peterjan:
This came up in a [`renterd` review](https://github.com/SiaFoundation/renterd/pull/264#discussion_r1160522122) the other day so I decided to propose it. I'm still not sure whether it's silly or actually useful so I figured I would let you be the judge Luke. Definitely feel free to just close this out, I was expecting way more occurrences to be honest, was surprised to only find a handful.


lukechampine:
yeah it's a borderline case. In `v2`, contracts have explicit renter and host fields -- no more slices -- so the constants will be unnecessary.

I think I'm ok with it, but let's call them something more generic (perhaps `Renter/HostContractIndex`), since the index is used for the unlock conditions too.

PR #111, opened 2023-04-18
Network Name

peterjan:
This PR adds a `Name` field to the `Network` object in the `consensus` package. By adding the network name the UI can differentiate between the `mainnet` and the `testnet` and adjust accordingly.


lukechampine:
Yeah this is definitely cleaner than inferring the name from other parameters.

To expose this to the UI, you'll want a `/consensus/network` endpoint that returns `cm.TipState().Network`, since by default the `Network` field is not encoded.

PR #113, opened 2023-04-19
Check the RPC costs for overflow

mike76-dev:
This should prevent an overflow panic when the hosts set their prices ridiculously high. Could fix renterd/#283

### DIFF
--- a/rhp/v2/rhp.go
+++ b/rhp/v2/rhp.go
@@ -282,9 +282,15 @@ type (
 
 // RPCSectorRootsCost returns the price of a SectorRoots RPC.
 func RPCSectorRootsCost(settings HostSettings, n uint64) types.Currency {
-	return settings.BaseRPCPrice.
-		Add(settings.DownloadBandwidthPrice.Mul64(n * 32)).  // roots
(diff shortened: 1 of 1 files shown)


peterjan:
Hi Mike, thank you for this contribution! 

I think when we added the overflow math to the `Currency` type we agreed we would not be performing these overflow checks all over the place but instead perform some rudimentary gouging checks first. If we find the host is gouging its prices, we simply stop entertaining the thought of using that host and consider it unusable. The host scoring should simply not execute these cost calculations if we find the host is price gouging, preventing this panic and keeping `core` clean.

I opened a PR in `renterd` [here](https://github.com/SiaFoundation/renterd/pull/291), it has to go through review still but I think we'll be able to close this out in favour of handling it in `renterd`.

lukechampine:
See previous discussion here: https://github.com/SiaFoundation/core/pull/89

Basically, anywhere we'd want to check `Currency` for overflow, we'd want to check other integer types as well. Given that there's no ergonomic way to do this, we settle for validating instead. I'm happy with that compromise for now.

PR #114, opened 2023-04-27
Add InitProgramCost helper 

ChrisSchinnerl:
This PR adds the `InitProgramHelper` which makes it a bit nicer to compute costs for programs.

e.g. computing the cost of a read sector program becomes this
```go
rc := pt.InitProgramCost()
rc.Add(pt.ReadSectorCost(rhpv2.SectorSize))
cost, _ := rc.Total()
```
where previously we couldn't make use of the `ResourceCost` type for the base cost.

Also snuck in a receiver change that Nate proposed.

### DIFF
--- a/rhp/v3/rhp.go
+++ b/rhp/v3/rhp.go
@@ -230,7 +230,7 @@ type (
 )
 
 // Add adds two ResourceCosts together.
-func (a *ResourceCost) Add(b ResourceCost) ResourceCost {
+func (a ResourceCost) Add(b ResourceCost) ResourceCost {
 	return ResourceCost{
 		Base:       a.Base.Add(b.Base),
 		Storage:    a.Storage.Add(b.Storage),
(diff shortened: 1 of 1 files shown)


rhp/v3/rhp.go
```diff
@@ -307,6 +307,13 @@ func (pt *HostPriceTable) HasSectorCost() ResourceCost {
 	}
 }
 
+// InitProgramCost is the cost of initialising an mdm program.
+func (pt *HostPriceTable) InitProgramCost() ResourceCost {
```
lukechampine, line 311: how about `BaseCost`?

ChrisSchinnerl: Ok, I updated it to `BaseCost`. Honestly any name is fine with me. I just want to get rid of 
```go
rc := pt.ReadSectorCost(...)
cost, _ := rc.Total()
cost = cost.Add(pt.InitBaseCost)
```

PR #116, opened 2023-05-05
Remove quotes from Specifier.String()

n8maninger:
```go
// MarshalText implements encoding.TextMarshaler.
func (uk UnlockKey) MarshalText() ([]byte, error) {
	return marshalHex(uk.Algorithm.String(), uk.Key[:])
}
```

Causes UnlockKey to JSON encode as `"\"ed25519\":48a468d62e48e27ef24707bd1f467476e9605954e201834c659e614b9704d041"`

This will likely break a bunch of things in `hostd` and `renterd`. Is adding compatibility code for this in `core` worth it?

### DIFF
--- a/types/types.go
+++ b/types/types.go
@@ -599,7 +599,7 @@ func ParseChainIndex(s string) (ci ChainIndex, err error) {
 }
 
 // String implements fmt.Stringer.
-func (s Specifier) String() string { return strconv.Quote(string(bytes.Trim(s[:], "\x00"))) }
+func (s Specifier) String() string { return string(bytes.Trim(s[:], "\x00")) }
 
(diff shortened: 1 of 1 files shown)


lukechampine:
dang.

Specifiers are annoying because they aren't always simple strings. For example, `siad` has a handful of specifiers that end in `\n`, as well as a few that end in `types.RuneToString(2)`. That's why I chose to quote them in the `String` method. However, I don't see any "annoying" specifiers in `core`, `renterd`, or `hostd`. So I agree that we should remove the quotes, and maybe add a check to `NewSpecifier` that rejects non-alphanumeric characters.

As for compatibility... yeah, we should make `UnmarshalText` handle both cases for a while. We can toss it once we're out of beta.

PR #117, opened 2023-05-24
Have SiamuxAddr return an empty string if no host was specified in the settings

ChrisSchinnerl:
To avoid accidentally dialing localhost in case the port is set but the address isn't.


lukechampine:
(nbd, but the commit prefix should be `rhp` here, not `v2`)

PR #120, opened 2023-06-02
Update PayByContract to return 'false' if a contract has reached the max revision number

ChrisSchinnerl:
Not sure if this is the right place to put this considering that the return type is a boolean, but some kind of safeguard against this seems appropriate.

### DIFF
--- a/rhp/v3/rhp.go
+++ b/rhp/v3/rhp.go
@@ -2,6 +2,7 @@
 package rhp
 
 import (
+	"math"
 	"time"
 
 	"go.sia.tech/core/consensus"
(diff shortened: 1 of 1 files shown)


lukechampine:
Having `false` represent multiple distinct types of failure is suspicious. Case in point, it looks like `renterd` currently does:
```go
        payment, ok := rhpv3.PayByContract(rev, amount, refundAcct, sk)
        if !ok {
                return rhpv3.PayByContractRequest{}, ErrInsufficientFunds
        }
```
which would be incorrect behavior if `!ok` can mean "contract not revisable." So this should probably return an error instead.

The other option would be to match the API of `PayByEphemeralAccount`: instead of failing, return a revision that the host will (presumably) reject. But I think it's better to catch errors locally when possible.

lukechampine:
@ChrisSchinnerl are you still interested in merging this, or is `renterd` using a different approach now?

PR #121, opened 2023-07-06
Refactor DB

lukechampine:
This began as an effort to make the DB commit less frequently, which is important for speeding up IBD. Currently, we commit (i.e. fsync) after every block, which means hundreds of thousands of fsyncs are required for a full IBD -- often many fsyncs per second. Needless to say, this is enormously wasteful. Ideally, we would fsync once every few seconds, and/or after buffering a large number of writes. So that's what this PR does: it tracks how long it's been since the last fsync, and how many writes we've buffered since then, and only flushes after we cross a threshold.

In the course of implementing this, I realized that our `DB` interface was more powerful than necessary. Specifically, it supports fully-isolated transactions, meaning you can have multiple "views" of the DB open concurrently, and those views won't observe each others effects, and can be committed or rolled back independently. However, our usage of the DB is already serialized by the `chain.Manager`'s mutex, and clients of `chain.Manager` only have limited access to the DB. So we actually don't need transactions; all we need is a way to force an fsync, so that we can guarantee that a subscriber will never commit data that hasn't been committed by the manager itself. That's what this PR does.

This still needs to be tested and benchmarked on an actual node, so I'll hold off on merging until then. Hopefully we'll see a significant speedup!

### DIFF
--- a/chain/db.go
+++ b/chain/db.go
@@ -14,14 +14,9 @@ import (
 
 // A DB is a generic key-value database.
 type DB interface {
-	Begin() (DBTx, error)
-}
-
-// A DBTx is a transaction executed on a key-value database.
-type DBTx interface {
 	Bucket(name []byte) DBBucket
 	CreateBucket(name []byte) (DBBucket, error)
-	Commit() error
+	Flush() error
 	Cancel()
 }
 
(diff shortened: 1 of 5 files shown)


chris124567:
Here are some benchmarks I did a few days ago here for any outside trying to follow this PR:

image.

In short any form of batching is substantially better than no batching.

PR #122, opened 2023-08-06
Reuse encoder in DB

chris124567:
Prevent an unnecessary 1 KB allocation from being repeated in all the `put` calls.

### DIFF
--- a/chain/db.go
+++ b/chain/db.go
@@ -176,9 +176,9 @@ func (b *dbBucket) putRaw(key, value []byte) {
 
 func (b *dbBucket) put(key []byte, v types.EncoderTo) {
 	var buf bytes.Buffer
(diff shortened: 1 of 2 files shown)


chain/db.go
```diff
@@ -623,7 +625,11 @@ func NewDBStore(db DB, n *consensus.Network, genesisBlock types.Block) (_ *DBSto
 		return nil, Checkpoint{}, errors.New("detected siad database, refusing to proceed")
 	}
 
-	dbs := &DBStore{db: db, n: n}
+	dbs := &DBStore{
+		db:  db,
+		n:   n,
+		enc: &types.Encoder{},
```
lukechampine, line 631: I think we can use a non-pointer for the `enc` field; that way, we don't need to explicitly initialize it here

PR #123, opened 2023-08-28
RPCCost (RHPv2)

peterjan:
This PR adds the `rpcCost` type from `hostd`. We can move it over to `core` and get rid of some of the very rough estimates for proof sizes and replace them by exact values we get from `DiffProofSize`.

### DIFF
--- a/rhp/v2/rhp.go
+++ b/rhp/v2/rhp.go
@@ -3,8 +3,8 @@ package rhp
 
 import (
 	"encoding/json"
+	"errors"
 	"fmt"
-	"math/bits"
 	"net"
 	"strings"
 	"time"
@@ -139,6 +139,29 @@ var (
 	RPCWriteActionUpdate = types.NewSpecifier("Update")
 
 	RPCReadStop = types.NewSpecifier("ReadStop")
+)
+
+// RPC read/write errors
+var (
+	// ErrOffsetOutOfBounds is returned when the offset exceeds and length
+	// exceed the sector size.
+	ErrOffsetOutOfBounds = errors.New("update section is out of bounds")
+
+	// ErrInvalidSectorLength is returned when a sector is not the correct
+	// length.
+	ErrInvalidSectorLength = errors.New("length of sector data must be exactly 4MiB")
+	// ErrSwapOutOfBounds is returned when one of the swap indices exceeds the
+	// total number of sectors
+	ErrSwapOutOfBounds = errors.New("swap index is out of bounds")
+	// ErrTrimOutOfBounds is returned when a trim operation exceeds the total
+	// number of sectors
+	ErrTrimOutOfBounds = errors.New("trim size exceeds number of sectors")
+	// ErrUpdateOutOfBounds is returned when the update index exceeds the total
+	// number of sectors
+	ErrUpdateOutOfBounds = errors.New("update index is out of bounds")
+	// ErrUpdateProofSize is returned when a proof is requested for an update
+	// operation that is not a multiple of 64 bytes.
+	ErrUpdateProofSize = errors.New("update section is not a multiple of the segment size")
 )
 
 // RPC request/response objects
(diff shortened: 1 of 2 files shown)


peterjan:
I will re-open a PR against `its-happening` 🤦‍♂️ , I never thought to check out the updated code there

lukechampine:
I'm not entirely sure why the RHP stuff in `v2` is different. In any case, the whole `v2` package is gone in `its-happening` -- I decided to take a different approach.

I think my preference is:
- Implement exact cost calculations in `core` now (for use in RHPv2/RHPv3)
- Use fixed costs in RHPv4

peterjan:
Yeah I saw you removed `v2` in your branch, that is why I thought you no longer wanted to use `DiffProofSize` and pass in the exact `RPCWriteAction`s to the functions calculating the cost of the RPC. I pushed a small commit that reverts my changes to the `v2` package and changes the function signatures a bit.

However, since you mention we ideally go for _exact_ cost calculations. Would you prefer changing some of those functions entirely to accept `[]RPCWriteAction` and implement `DiffProofSize` as you did in the `v2` package? I might go ahead and try to implement that already to see the difference.

peterjan:
For rhpv3 we could/should update the cost methods on the `HostPriceTable` to take an argument that indicates whether a proof is required and then also all necessary information to calculate the proof size. Is that something we want to add as well?

rhp/v2/rhp.go
```diff
@@ -280,45 +303,116 @@ type (
...
 
-// RPCAppendCost returns the price and collateral of a Write RPC with a single
-// append operation.
-func RPCAppendCost(settings HostSettings, storageDuration uint64) (price, collateral types.Currency) {
-	price = settings.BaseRPCPrice.
-		Add(settings.StoragePrice.Mul64(SectorSize).Mul64(storageDuration)).
-		Add(settings.UploadBandwidthPrice.Mul64(SectorSize)).
-		Add(settings.DownloadBandwidthPrice.Mul64(128 * 32)) // proof
-	collateral = settings.Collateral.Mul64(SectorSize).Mul64(storageDuration)
-	// add some leeway to reduce chance of host rejecting
```
peterjan, line 313: @n8maninger `hostd` stopped doing it so I copied that behaviour, just wanted to draw attention to it to ensure it's safe to do so

rhp/v2/rhp.go
```diff
@@ -280,45 +303,116 @@ type (
 	}
 )
 
-// RPCSectorRootsCost returns the price of a SectorRoots RPC.
-func RPCSectorRootsCost(settings HostSettings, n uint64) types.Currency {
-	return settings.BaseRPCPrice.
-		Add(settings.DownloadBandwidthPrice.Mul64(n * 32)).  // roots
-		Add(settings.DownloadBandwidthPrice.Mul64(128 * 32)) // proof
+type RPCCost struct {
```
peterjan, line 306: Functionally this type matches the `ResourceCost` in RHPv3, we could move this into the `types` package maybe and use the one type. I didn't want to do that yet mostly because of the documentation in `ResourceCost` which is tailored to the MDM... Maybe it's fine to just have both, for the same reason `hostd` had its own internal `rpcCost` type.

rhp/v2/rhp.go
```diff
@@ -280,45 +303,116 @@ type (
...
+
+func (c RPCCost) Total() (cost, collateral types.Currency) {
+	return c.Base.Add(c.Storage).Add(c.Ingress).Add(c.Egress), c.Collateral
 }
 
-// RPCReadCost returns the price of a Read RPC.
-func RPCReadCost(settings HostSettings, sections []RPCReadRequestSection) types.Currency {
-	sectorAccessPrice := settings.SectorAccessPrice.Mul64(uint64(len(sections)))
+// RPCReadCost returns the cost of a Read RPC.
+func RPCReadCost(sections []RPCReadRequestSection, proof bool, settings HostSettings) (RPCCost, error) {
```
n8maninger, line 329: Should these be moved onto `HostSettings` as we did for the price table in RHP3?

n8maninger (approved):

rhp/v2/rhp.go
```diff
@@ -280,45 +303,116 @@ type (
...
 	for _, sec := range sections {
-		bandwidth += sec.Length
-		bandwidth += 2 * uint64(bits.Len64(LeavesPerSector)) * 32 // proof
+		switch {
+		case uint64(sec.Offset)+uint64(sec.Length) > SectorSize:
+			return RPCCost{}, ErrOffsetOutOfBounds
+		case sec.Length == 0:
+			return RPCCost{}, errors.New("length cannot be zero")
+		case proof && (sec.Offset%LeafSize != 0 || sec.Length%LeafSize != 0):
+			return RPCCost{}, errors.New("offset and length must be multiples of SegmentSize when requesting a Merkle proof")
```
ChrisSchinnerl, line 339: *LeafSize

ChrisSchinnerl (approved):

PR #125, opened 2023-09-21
More tests for its-happening

chris124567:
Add a test similar to TestP2P in walletd and add some extra cases mostly relating to file contracts to the validation test.

### DIFF
--- a/consensus/validation_test.go
+++ b/consensus/validation_test.go
@@ -2,6 +2,7 @@ package consensus_test
 
 import (
 	"bytes"
+	"math"
 	"testing"
 	"time"
 
@@ -141,6 +142,7 @@ func TestValidateBlock(t *testing.T) {
 		t.Fatal(err)
 	}
 
+	// tests with correct signatures
 	{
 		tests := []struct {
 			desc    string
@@ -195,6 +197,16 @@ func TestValidateBlock(t *testing.T) {
 				"overflowing miner payout",
 				func(b *types.Block) {
 					b.MinerPayouts = []types.SiacoinOutput{
+						{Address: types.VoidAddress, Value: types.MaxCurrency},
+						{Address: types.VoidAddress, Value: types.MaxCurrency},
+					}
+				},
+			},
+			{
+				"overflowing siacoin outputs",
+				func(b *types.Block) {
+					txn := &b.Transactions[0]
+					txn.SiacoinOutputs = []types.SiacoinOutput{
 						{Address: types.VoidAddress, Value: types.MaxCurrency},
 						{Address: types.VoidAddress, Value: types.MaxCurrency},
 					}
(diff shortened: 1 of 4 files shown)


chain/p2p_test.go
```diff
@@ -0,0 +1,392 @@
...
+	return nil, nil, nil
+}
+func (h handler) Checkpoint(types.ChainIndex) (types.Block, consensus.State, error) {
+	return types.Block{}, consensus.State{}, nil
+}
+func (h handler) RelayV2Header(gateway.V2BlockHeader, *gateway.Peer)         {}
+func (h handler) RelayV2BlockOutline(gateway.V2BlockOutline, *gateway.Peer)  {}
+func (h handler) RelayV2TransactionSet([]types.V2Transaction, *gateway.Peer) {}
+
+func TestP2P(t *testing.T) {
```
lukechampine, line 104: so the bad news is that this test only adds 1% of coverage to `chain`. Not worth it imo. I know it was a pain to get it working, sorry :(

The good news is that this test *will* be nice to have if we end up moving more syncer code into `core`. So we should keep it around in the git history for that, at least.

I think the best place to focus on for now is file contract tests in `consensus` (v1 and v2). It's crucial that we test as many edge cases as possible in the new consensus code, so we should strive for 100% coverage.

consensus/validation_test.go
```diff
@@ -473,6 +473,22 @@ func TestValidateV2Block(t *testing.T) {
...
+				r.RenterSignature = renterPrivateKey.SignHash(cs.RenewalSigHash(*r))
+				r.HostSignature = hostPrivateKey.SignHash(cs.RenewalSigHash(*r))
+				txn.FileContractResolutions[i].Resolution = r
+				break
+			case *types.V2FileContractFinalization:
+				fc := types.V2FileContract(*r)
+				fc.RenterSignature = renterPrivateKey.SignHash(cs.ContractSigHash(fc))
+				fc.HostSignature = hostPrivateKey.SignHash(cs.ContractSigHash(fc))
+
+				txn.FileContractResolutions[i].Resolution = (*types.V2FileContractFinalization)(&fc)
```
lukechampine, line 488: since `r` is a pointer, you don't need to set `txn.FileContractResolutions[i].Resolution = r`. For the `types.V2FileContractFinalization` branch, I would do:
```go
sigHash := cs.ContractSigHash(types.V2FileContract(*r))
r.RenterSignature = renterPrivateKey.SignHash(sigHash)
r.HostSignature = hostPrivateKey.SignHash(sigHash)
```

types/encoding.go
```diff
@@ -598,6 +598,10 @@ func (ren V2FileContractRenewal) EncodeTo(e *Encoder) {
 	ren.HostSignature.EncodeTo(e)
 }
 
+func (fcf V2FileContractFinalization) EncodeTo(e *Encoder) {
```
lukechampine, line 601: docstring

consensus/validation_test.go
```diff
@@ -428,16 +470,82 @@ func TestValidateBlock(t *testing.T) {
...
+
+			if err := consensus.ValidateBlock(cs, corruptBlock, dbStore.SupplementTipBlock(corruptBlock)); err == nil {
+				t.Fatalf("accepted block with %v", test.desc)
+			}
+		}
+	}
+}
+
+func updateProofs(cau consensus.ApplyUpdate, sces []types.SiacoinElement, sfes []types.SiafundElement, fces []types.V2FileContractElement) {
+	for i := 0; i < len(sces); i++ {
```
lukechampine, line 529: these can be `for i := range sces`

lukechampine:
We're above 75% coverage in consensus now, nice!

PR #134, opened 2023-12-06
Add golangci-lint

chris124567:
I have been adding this to our repos so we can do things like check struct tags automatically.  I know core is still changing so if this is unwanted right now please feel free to close this PR.

Edit: Need to do some Ledger review fixes today.  Will fix remaining warnings tomorrow.


consensus/state.go
```diff
@@ -71,7 +71,7 @@ type Network struct {
 		Height    uint64        `json:"height"`
 		OakTime   time.Duration `json:"oakTime"`
 		OakTarget types.BlockID `json:"oakTarget"`
-	} `json:"hardforkASIC"`
+	} `json:"hardforkAsic"`
```
lukechampine, line 74: definitely not. ASIC is an acronym.

lukechampine:
Let's disable `modifies-value-receiver` and `indent-error-flow`. `if-return` is stupid too (it makes refactoring more annoying) but I guess it's situational, and it only comes up in `smux` which I don't care about.

consensus/state.go
```diff
@@ -67,6 +67,7 @@ type Network struct {
 		FixHeight        uint64    `json:"fixHeight"`
 		GenesisTimestamp time.Time `json:"genesisTimestamp"`
 	} `json:"hardforkOak"`
+	//nolint:tagliatelle
```
lukechampine, line 70: unfortunate. There's an issue open for specifying custom initialisms in tagliatelle, but no indication that it'll be implemented any time soon.

lukechampine: btw, old tag name needs to be restored

chris124567: Sorry about that... Didn't sleep well last night

n8maninger:
IIRC, we fixed the CI fail on Windows by adding the following step:
```yml
- name: Configure git # required for golangci-lint on Windows
  run: git config --global core.autocrlf false
```

PR #136, opened 2023-12-11
Add `RenewalCosts` to rhpv3 

ChrisSchinnerl:
We ran into bugs related to determining the right amount of collateral when renewing contracts. Most notably exceeding the max collateral setting of hosts.

There were multiple reasons for that. I was able to get rid of most errors by postponing computing the collateral until after we have acquired a contract.

The other issue, which is the reason for this PR, is the fact that we used `ContractRenewalCollateral` to compute the `newCollateral` and then `CalculateHostPayouts` to create the payouts and both of these functions perform the same math but slightly different. (one uses fc.EndHeight() and the other one fc.WindowEnd` for determining the extension of the renewal).

Since we don't use rhpv2 renewals I'm not going to bother refactoring that but I'm updating rhpv3 to make sure there is no duplicate code anymore and that everything, including the capping of the `collateral` according to the host's `MaxCollateral` happens in one place.

Right now this is deployed on Arequipa and there are no more collateral errors.

### DIFF
--- a/rhp/v2/contracts.go
+++ b/rhp/v2/contracts.go
@@ -81,7 +81,7 @@ func ContractRenewalCollateral(fc types.FileContract, expectedNewStorage uint64,
 	if endHeight < fc.EndHeight() {
 		panic("endHeight should be at least the current end height of the contract")
 	}
-	extension := endHeight - fc.EndHeight()
+	extension := endHeight - fc.WindowEnd
 	if endHeight < blockHeight {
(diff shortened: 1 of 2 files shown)


lukechampine:
As someone who has painstakingly stepped through the renewal collateral code multiple times, and still (apparently) gotten it wrong -- my condolences. It is not fun to painstakingly step through the renewal collateral code.

PR #137, opened 2024-01-14
Add Siacoins() to Currency

n8maninger:


n8maninger:
Used by https://github.com/SiaFoundation/hostd/pull/273

types/currency.go
```diff
@@ -276,6 +278,20 @@ func (c Currency) String() string {
 	return s
 }
 
+// Float64 returns the value of c as a float64 and a bool indicating whether f
+// represents x exactly.
+func (c Currency) Float64() (float64, bool) {
+	return new(big.Rat).SetFrac(c.Big(), scPrecisionBig).Float64()
+}
```
lukechampine, line 285: If this returns the number of SC, it should be called `Siacoins()`, not `Float64`. And I would drop the bool; when would a caller care about whether the representation is exact?

Also I would use `HastingsPerSiacoin.Big()` instead of defining a separate variable for it

PR #140, opened 2024-02-18
Add tests for ElementAccumulator encoding, StorageProofRoot calculation, and ElementProof update logic

new0nebit:
A mysterious force of nature inspired me to write this code today, thereby increasing the code coverage to 75.7%


consensus/merkle_test.go
```diff
@@ -37,3 +42,99 @@ func TestElementAccumulatorEncoding(t *testing.T) {
...
+
+            if tt.expectPanic {
+                defer func() {
+                    if r := recover(); r == nil {
+                        t.Errorf("The code did not panic for %s", tt.name)
+                    }
+                }()
+            }
+
+            eru.updateElementProof(e)
```
lukechampine, line 131: try this:
```go
var didPanic bool
func() {
    defer func() { didPanic = recover() != nil }()
    eru.updateElementProof(e)
}()
```

consensus/merkle_test.go
```diff
@@ -37,3 +42,99 @@ func TestElementAccumulatorEncoding(t *testing.T) {
...
+    }{
+        {
+            name:      "ValidProof",
+            leafHash:  leafHash,
+            leafIndex: 0,
+            filesize:  1024,
+            proof:     proof,
+            wantRoot:  types.Hash256{},
+            wantErr:   false,
+        },
```
lukechampine, line 70: this should be named `"InvalidProof"` :P 

Let's have one valid test case and two invalid cases. For the valid case, use a filesize that isn't a perfect power of 2 (how about `829`) and leaf index somewhere in the middle of tree (how about `10`). For those values, the proof slice should have a length of `4`. When you run the test, the root will be some random-looking value; copy that into the test case as `wantRoot`.  For the invalid cases, use the same fields as the valid case, but make the proof longer or shorter. You can get rid of the `wantErr` field.

new0nebit: Thanks! I spent a lot of time trying to cover this function. However, using your suggested approach I made corrections, added new test cases and finally managed to cover it! Code coverage has now reached 75.9% I would be glad to receive your additional comments or suggestions for improvement!

consensus/merkle_test.go
```diff
@@ -45,36 +45,66 @@ func TestElementAccumulatorEncoding(t *testing.T) {
...
+    validProof := []types.Hash256{
+        {0x0A, 0x0B, 0x0C},
+        {0x0D, 0x0E, 0x0F},
+        {0x10, 0x11, 0x12},
+        {0x13, 0x14, 0x15},
     }
+    longProof := append(validProof, types.Hash256{0x16, 0x17, 0x18})
+    shortProof := validProof[:2]
+
+    validWantRoot := types.Hash256{0xa8, 0x9d, 0xbb, 0xb5, 0x45, 0xaa, 0x4b, 0x46, 0x69, 0x62, 0x30, 0xd1, 0x04, 0x07, 0x6f, 0x06, 0xb5, 0x7a, 0x6a, 0xb0, 0x8b, 0x23, 0x41, 0xc3, 0xd0, 0x12, 0xbd, 0xc1, 0x3e, 0x23, 0xeb, 0x35}
```
lukechampine, line 57: you can make this more compact with `UnmarshalText`:
```go
var validWantRoot types.Hash256
validWantRoot.UnmarshalText([]byte(`h:a89dbbb545aa4b46696230d104076f06b57a6ab08b2341c3d012bdc13e23eb35`))
```

new0nebit: Would it be right to also add an error message to report that the hash does not match the expected format?

```go
var validWantRoot types.Hash256
    err := validWantRoot.UnmarshalText([]byte("a89dbbb545aa4b46696230d104076f06b57a6ab08b2341c3d012bdc13e23eb35"))
    if err != nil {
        t.Fatalf("Error deserializing the textual representation of the storage proof root hash: %v", err)
    }

Comment (lukechampine) [reply consensus/merkle_test.go:57]: In this case it's not really necessary. If the `UnmarshalText` fails, `validWantRoot` will be all zeros and the test will still fail.

File: consensus/merkle_test.go [hunk 4]
```diff
@@ -37,3 +42,130 @@ func TestElementAccumulatorEncoding(t *testing.T) {
...
+
+    for _, tt := range tests {
+        t.Run(tt.name, func(t *testing.T) {
+            gotRoot := storageProofRoot(tt.leafHash, tt.leafIndex, tt.filesize, tt.proof)
+            if tt.valid {
+                if !bytes.Equal(gotRoot[:], tt.wantRoot[:]) {
+                    t.Errorf("%s failed: got %v, want %v", tt.name, gotRoot, tt.wantRoot)
+                }
+            } else {
+                if bytes.Equal(gotRoot[:], tt.wantRoot[:]) {
```
lukechampine, line 106: arrays can be compared directly with `==`: `gotRoot == tt.wantRoot`

lukechampine (changes requested):
looks like the linter is complaining 🤔

new0nebit:
> looks like the linter is complaining 🤔

Should be fixed now! 🫡

All checks have passed

lukechampine (approved):

lukechampine:
Thanks!

PR #141, opened 2024-02-27
Update revision when revising a FileContractElement that's found in the ephemeral cache

peterjan:
When mining a block with transactions:
- contract formation
- contract revision (broadcast)
- contract revision (broadcast)
- contract revision (broadcast)

I noticed my subscriber receives only 3 file contract elements in the ApplyUpdate, more importantly though one of those did not have a revision. I think there's a bug in reviseFileContractElement that doesn't properly update the `rev` map when encountering an `fce` we find in `ephemeral`, which I assume is a sort of cache.


peterjan:
expected behaviour after all - closing

PR #146, opened 2024-03-11
types: Make (Currency).String exact

lukechampine:
```
Old: "~2.529 mS"
New: "2.529378333356156158367 mS"
```
We discussed removing the SI units as well, and using SC exclusively (i.e. the example above would be `2.529378333356156158367 mS`), but there was less consensus on that.

Since `String` is now exact, `ExactString` has been removed; if you want raw Hastings values, you can use `Big` or `fmt.Sprintf("%d", c)`. If you want a rounded value, you can use standard approaches to round the `float64` returned by `Siacoins`.

### DIFF
--- a/types/currency.go
+++ b/types/currency.go
@@ -228,56 +228,31 @@ func (c Currency) Big() *big.Int {
 
 // ExactString returns the base-10 representation of c as a string.
 func (c Currency) ExactString() string {
-	if c.IsZero() {
-		return "0"
-	}
-	buf := []byte("0000000000000000000000000000000000000000") // log10(2^128) < 40
-	for i := len(buf); ; i -= 19 {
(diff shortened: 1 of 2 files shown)


n8maninger:
I'm okay with the change to `String()`, but I'm against the removal of `ExactString()`. 

Personally, I'd rather have `String()` return the value in hastings. All code works with hastings and a `Currency` represents a number of hastings not Siacoin.

lukechampine:
fair enough, restored.

I've also tweaked `UnmarshalText` to accept strings with units. This not only benefits weirdos like me who write JSON transactions by hand, but also anyone using Currency values in a human-readable config format (e.g. a YAML file specifying host prices? idk)

ChrisSchinnerl (approved):
Got a bit nervous so I pushed a small commit to cover `UnmarshalText`.

n8maninger (approved):

PR #148, opened 2024-03-22
Add Test Coverage for SatisfiedPolicy and enhance SpendPolicy tests

new0nebit:
I started writing this code during the last hackathon for fun, practicing writing test cases. But I didn't finish it because I had to switch to the main task regarding consensus coverage. I felt bad for wipe out a work that had been done, so I ended up finishing the code and also making some changes to it. Initially, I was only aiming to cover the `SatisfiedPolicy` but got carried away and started covering the `PolicyTypeAbove` and `PolicyTypePublicKey`, completely overlooking the fact that you had already done this 😅 Anyway, here it is. I hope that this alternative approach and the additional scenarios will prove valuable to you 🙏

### DIFF
--- a/types/policy_test.go
+++ b/types/policy_test.go
@@ -2,11 +2,9 @@ package types
 
 import (
 	"bytes"
-	"crypto/sha256"
-	"encoding/hex"
 	"encoding/json"
 	"fmt"
-	"strconv"
+	"reflect"
 	"testing"
 	"time"
 )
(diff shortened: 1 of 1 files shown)


lukechampine:
Thanks for contributing more tests! I'll give this a proper review in the morning, but from a first pass it looks fine.

types/policy_test.go
```diff
@@ -329,3 +333,123 @@ func TestPolicyRoundtrip(t *testing.T) {
...
+func TestPolicyMarshaling(t *testing.T) {
+	privateKey := GeneratePrivateKey()
+	publicKey := privateKey.PublicKey()
+	publicKeyHex := hex.EncodeToString(publicKey[:])
+
+	// Generate a unique SHA256 hash from the current Unix time in nanoseconds for PolicyTypeHash test.
+	currentTime := time.Now().UnixNano()
+	currentTimeBytes := []byte(strconv.FormatInt(currentTime, 10))
+	hash := sha256.Sum256(currentTimeBytes)
+	hashHex := hex.EncodeToString(hash[:])
```
lukechampine, line 346: these values don't need to be random; you can do:
```go
publicKey := NewPrivateKeyFromSeed(make([]byte, 32)).PublicKey()
hash := HashBytes(nil)
```
This ensures that the test will be deterministic, so the result can be cached.

types/policy_test.go
```diff
@@ -329,3 +333,123 @@ func TestPolicyRoundtrip(t *testing.T) {
...
+
+	tests := []struct {
+		name        string
+		input       interface{}
+		output      string
+		marshalFunc func(interface{}) ([]byte, error)
```
lukechampine, line 352: how about:
```go
struct {
    sp  SpendPolicy
    exp string
}
```

types/policy_test.go
```diff
@@ -329,3 +333,123 @@ func TestPolicyRoundtrip(t *testing.T) {
...
+	}{
+		{
+			name:   "MarshalText",
+			input:  SpendPolicy{Type: PolicyTypeAbove(100)},
```
lukechampine, line 356: you can use the various SpendPolicy constructor functions for these:
```go
sp: PolicyAbove(100)

sp: PolicyAfter(time.Unix(1234567890, 0))
```
etc.

types/policy_test.go
```diff
@@ -329,3 +333,123 @@ func TestPolicyRoundtrip(t *testing.T) {
...
+			input:  SpendPolicy{Type: PolicyTypeAfter(time.Unix(1234567890, 0))},
+			output: "after(1234567890)",
+			marshalFunc: func(v interface{}) ([]byte, error) {
+				return v.(SpendPolicy).MarshalText()
+			},
+		},
+		{
+			name:   "MarshalText with PolicyTypePublicKey",
+			input:  SpendPolicy{Type: PolicyTypePublicKey(publicKey)},
+			output: "pk(0x" + publicKeyHex + ")",
```
lukechampine, line 373: `fmt.Sprintf("pk(0x%x)", publicKey[:])`

types/policy_test.go
```diff
@@ -329,3 +333,123 @@ func TestPolicyRoundtrip(t *testing.T) {
...
+			marshalFunc: func(v interface{}) ([]byte, error) {
+				return v.(SpendPolicy).MarshalText()
+			},
+		},
+		{
+			name:   "MarshalText with PolicyTypeHash",
+			input:  SpendPolicy{Type: PolicyTypeHash(hash)},
+			output: "h(0x" + hashHex + ")",
```
lukechampine, line 381: ditto

types/policy_test.go
```diff
@@ -329,3 +333,123 @@ func TestPolicyRoundtrip(t *testing.T) {
...
+			},
+		},
+		{
+			name: "MarshalJSON",
+			input: SatisfiedPolicy{
+				Preimages: [][]byte{{0xde, 0xad, 0xbe, 0xef}, {0xba, 0xad, 0xf0, 0x0d}},
+			},
+			output:      "preimages",
+			marshalFunc: json.Marshal,
+		},
```
lukechampine, line 393: This test case isn't like the others. I suggest removing it for now. We can add a separate test for `SatisfiedPolicy` later.

types/policy_test.go
```diff
@@ -329,3 +333,123 @@ func TestPolicyRoundtrip(t *testing.T) {
...
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
```
lukechampine, line 397: Subtests are kinda overkill here.

types/policy_test.go
```diff
@@ -329,3 +333,123 @@ func TestPolicyRoundtrip(t *testing.T) {
...
+			data, err := tt.marshalFunc(tt.input)
```
lukechampine, line 398: `json.Marshal` should work fine

types/policy_test.go
```diff
@@ -329,3 +333,123 @@ func TestPolicyRoundtrip(t *testing.T) {
...
+			if err != nil {
+				t.Fatalf("Expected no error, but got %v", err)
+			}
+
+			if !bytes.Contains(data, []byte(tt.output)) {
```
lukechampine, line 403: I would prefer an exact match, like `string(data) != tt.exp`. You'll want to surround the test case strings with backticks, e.g.:
```go
exp: `"above(100)"`,
```

types/policy_test.go
```diff
@@ -329,3 +333,123 @@ func TestPolicyRoundtrip(t *testing.T) {
...
+		{
+			name:      "InvalidPolicy",
+			jsonData:  `{"Policy": "ShouldBeAnObjectOrValidType", "Signatures": null, "Preimages": []}`,
+			expectErr: true,
+		},
+		{
+			name:          "ValidPreimage",
+			jsonData:      `{"Policy": null, "Signatures": null, "Preimages": ["68656c6c6f776f726c64"]}`,
+			expectErr:     false,
+			preimage:      "helloworld",
```
lukechampine, line 432: this doesn't appear to be used

types/policy_test.go
```diff
@@ -329,3 +333,123 @@ func TestPolicyRoundtrip(t *testing.T) {
...
+
+			if tt.checkPreimage {
+				expectedPreimage, err := hex.DecodeString("68656c6c6f776f726c64")
+				if err != nil {
+					t.Fatalf("hex.DecodeString error: %v", err)
+				}
+				if !bytes.Equal(sp.Preimages[0], expectedPreimage) {
+					t.Errorf("Preimage mismatch")
+				}
+			}
```
lukechampine, line 452: I think what you want here is:
```go
if len(tt.preimages) != 0 && !reflect.DeepEqual(tt.preimages, sp.Preimages) {
    t.Error("preimage mismatch")
}
```
And then, for your preimage test case, do:
```go
preimages: [][]byte{[]byte("helloworld")},
```

lukechampine (changes requested):

lukechampine (approved):

PR #163, opened 2024-05-29
Add maxOutputLength to limit slice creation in RPCExecuteProgramResponse

ChrisSchinnerl:
Something crashed Arequipa by returning a huge `OutputLength`. 10MiB should be plenty.

### DIFF
--- a/rhp/v3/encoding.go
+++ b/rhp/v3/encoding.go
@@ -10,10 +10,6 @@ import (
 	"go.sia.tech/core/types"
 )
 
-const (
-	maxOutputLength = 10 * 1 << 20 // 10 MiB
-)
-
 // EncodeTo implements ProtocolObject.
 func (r *RPCError) EncodeTo(e *types.Encoder) {
 	r.Type.EncodeTo(e)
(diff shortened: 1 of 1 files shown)


rhp/v3/encoding.go
```diff
@@ -323,6 +327,10 @@ func (r *RPCExecuteProgramResponse) DecodeFrom(d *types.Decoder) {
 	}
 	(*types.V1Currency)(&r.TotalCost).DecodeFrom(d)
 	(*types.V1Currency)(&r.FailureRefund).DecodeFrom(d)
+	if r.OutputLength > maxOutputLength {
+		d.SetErr(fmt.Errorf("output length %d exceeds maximum of 10 MiB", r.OutputLength))
+		return
+	}
```
lukechampine, line 333: I wonder if you could do `r.OutputLength = uint64(d.ReadPrefix())`? That would automatically limit the size to the number of undecoded bytes remaining.

rhp/v3/encoding.go
```diff
@@ -323,6 +323,9 @@ func (r *RPCExecuteProgramResponse) DecodeFrom(d *types.Decoder) {
 	}
 	(*types.V1Currency)(&r.TotalCost).DecodeFrom(d)
 	(*types.V1Currency)(&r.FailureRefund).DecodeFrom(d)
+	if d.Err() != nil {
+		return // abort before reading output
```
peterjan, line 327: The way `SetErr` works it's not necessary to return early here because it clears the buffer. Coincidentally we also check for `err == nil` in the `for` loop in `d.Read`. We save the allocation but I'd still remove it.

peterjan (approved):

lukechampine:
from +8 to +1/-1, you love to see it

PR #166, opened 2024-06-06
Clear decoder buffer

n8maninger:
Ensures the decoder's buffer is cleared when `Read` encounters an error. This should fix OOM issues we're seeing when syncing Mainnet. It may also fix https://github.com/SiaFoundation/hostd/issues/406, but I haven't looked deep into that one yet.


n8maninger:
@lukechampine I think there's another potential attack vector where a huge prefix for an array is sent that is technically smaller than `d.lr.N`. since we use `make([]foo, d.ReadPrefix())` everywhere the decoder allocates the entire slice immediately which could expand significantly. 

For example, when decoding a block sent by a peer, `d.lr.N` is 50MB. That would enable a slice like `block.FileContracts` to allocate up to 50M items. I propose changing `ReadPrefix` to `ReadPrefix(max int64)` to prevent that.

types/encoding.go
```diff
@@ -170,12 +181,12 @@ func (d *Decoder) ReadUint64() uint64 {
 }
 
 // ReadPrefix reads a length prefix from the underlying stream. If the length
-// exceeds the number of bytes remaining in the stream, ReadPrefix sets d.Err
+// exceeds the maximum number of elements, ReadPrefix sets d.Err
 // and returns 0.
-func (d *Decoder) ReadPrefix() int {
+func (d *Decoder) ReadPrefix(limit int64) int {
```
n8maninger, line 186: All of these should be reviewed carefully to ensure we're not going to break existing clients.

types/encoding.go
```diff
@@ -136,8 +140,15 @@ func (d *Decoder) Read(p []byte) (int, error) {
 		if want > len(d.buf) {
 			want = len(d.buf)
 		}
-		var read int
-		read, d.err = io.ReadFull(&d.lr, d.buf[:want])
```
n8maninger, line 140: Since the buffer wasn't cleared when a read error happened, future calls to `ReadPrefix` would LE decode whatever was in the buffer. Occasionally a massive value would be read in the wrong spot causing massive allocations and OOM.

types/encoding.go
```diff
@@ -136,8 +136,15 @@ func (d *Decoder) Read(p []byte) (int, error) {
...
+		read, err := io.ReadFull(&d.lr, d.buf[:want])
+		if err != nil {
+			// When the decoder encounters an error, it must clear its buffer and
+			// return empty values for the remaining decodes. Because the decoder uses
+			// sticky errors instead of immediately returning a bad decode can
+			// potentially cause a massive allocation.
+			d.SetErr(err)
+			return n, err
+		}
```
lukechampine, line 151: this could also be written as:
```go
read, err := io.ReadFull(&d.lr, d.buf[:want])
n += copy(p[n:], d.buf[:read])
d.SetErr(err)
```
which feels a bit more in line with the "unconditional" style of the decoder. It also handles any trailing bytes that were read (though in practice this is unlikely to matter).

hmm, if we upgrade to go 1.21 then we can make it even more concise:
```go
func (d *Decoder) Read(p []byte) (int, error) {
	n := 0
	for len(p[n:]) > 0 && d.err == nil {
		read, err := io.ReadFull(&d.lr, d.buf[:min(len(p[n:]), len(d.buf))])
		n += copy(p[n:], d.buf[:read])
		d.SetErr(err)
	}
	return n, d.err
}
```
tempting! 😅

types/encoding.go
ChrisSchinnerl: Ahhh I see. That is quite unfortunate^^

peterjan: Is it that crazy to not pre-allocate the slice but grow it using `append`? Attackers could abuse the `limit` field and pass the maximum allowed for every type or situation right?

n8maninger: append wouldn’t help since the decoder doesn’t stop decoding. It returns the zero value for anything after the error

PR #170, opened 2024-06-11
types: Implement generic encoder functions

lukechampine:
Adds `EncodeSlice`, `EncodeSliceCast`, `EncodeSliceFn`, and their decoder counterparts, as well as `Cast` methods for increased type safety. This makes most encoding methods more succinct, but the real motivation was that it allows us to centralized the handling of malicious slice lengths (see discussion on #166). I will grudgingly admit that generics solved a real problem here. 😝

I'm not crazy about the API; it would be nicer if we could somehow unify these three functions, and if we could ditch the `Cast` methods. I explored replacing `EncodeSliceFn` with  a `BasicEncodable` type (wrapping `uint64`, `bool`, etc.), but it was kinda gross, requiring a run-time type-switch on every encode call.

I ran one manual benchmark to see whether this had a big performance impact, and it looked ok. We should benchmark it more thoroughly though.


types/encoding.go
```diff
@@ -210,6 +245,71 @@ type DecoderFunc func(*Decoder)
...
+		*v = new(T)
+		TP(*v).DecodeFrom(d)
+	} else {
+		*v = nil
+	}
+}
+
+// DecodeSlice decodes a length-prefixed slice of type T, containing values read
+// from the decoder.
+func DecodeSlice[T any, DF interface {
```
ChrisSchinnerl, line 263: Do we really need 2 generic parameters here? That is a shame.

types/encoding_test.go
ChrisSchinnerl, line 1: Any chance we can have some basic testing? ^^

Like a happy case test and making sure decoding random data doesn't panic?

types/encoding.go
lukechampine: as far as I can tell, yeah. They both get inferred, though ‾\\\_(ツ)\_/‾

n8maninger (approved):

PR #171, opened 2024-06-13
Add SpendPolicy fuzzer

chris124567:
Fuzz decoding of SpendPolicy which is probably has the most complicated decoding logic of all the types.  I have been running for ~30 minutes and coverage has not flatlined yet.

Run with:
```
$ go test -fuzz FuzzSpendPolicy ./types
```

### DIFF
--- a/types/encoding.go
+++ b/types/encoding.go
@@ -1186,6 +1186,10 @@ func (sp *SatisfiedPolicy) DecodeFrom(d *Decoder) {
 			}
 		case PolicyTypeUnlockConditions:
 			for i := range p.PublicKeys {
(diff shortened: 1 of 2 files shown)


types/encoding.go
```diff
@@ -1186,6 +1186,10 @@ func (sp *SatisfiedPolicy) DecodeFrom(d *Decoder) {
 			}
 		case PolicyTypeUnlockConditions:
 			for i := range p.PublicKeys {
+				if len(p.PublicKeys[i].Key) != 32 {
```
n8maninger, line 1189: Check if Algorithm is `SpecifierEd25519` as well

PR #180, opened 2024-07-09
Add JSON encoding for consensus updates

n8maninger:


lukechampine:
What's the use case? These types aren't intended to be serialized or stored. Unless you're writing consensus code, you should only see them inside a `ProcessChainUpdate` method.

n8maninger:
My goal was to get chain scanning working over HTTP. Serializing these types is the most straightforward way.

lukechampine:
Ah, I see. You basically want an HTTP route that wraps `(chain.Manager).UpdatesSince`. That makes sense. To be honest, I was thinking we were still using the old callback-based `chain.Subscriber` API. 😅

I think I'm onboard with the general idea of sending updates over HTTP.  The alternative would be sending blocks and states, and having the client call `consensus.ApplyBlock` and `consensus.RevertBlock`. But that would make a non-Go client even harder to write, since you'd have to reimplement a lot more consensus code. It does feel a bit wasteful to encode the entire `MidState`, since we don't care about a lot of its fields. I would be more inclined to approve this if we limited it to four new methods (`Marshal/Unmarshal` on `Apply/Revert`) and tightly scoped each of them. I'll try poking at this a bit.

PR #183, opened 2024-07-17
V2FileContractResolution serialization consistency 

Alrighttt:
This field does not follow the consistent use of camelCase for JSON serialization.

### DIFF
--- a/types/types.go
+++ b/types/types.go
@@ -1094,7 +1094,7 @@ func (res V2FileContractResolution) MarshalJSON() ([]byte, error) {
 	case *V2FileContractRenewal:
 		typ = "renewal"
 	case *V2StorageProof:
-		typ = "storage proof"
+		typ = "storageProof"
 	case *V2FileContractFinalization:
 		typ = "finalization"
 	case *V2FileContractExpiration:
@@ -1122,7 +1122,7 @@ func (res *V2FileContractResolution) UnmarshalJSON(b []byte) error {
 	switch p.Type {
 	case "renewal":
 		res.Resolution = new(V2FileContractRenewal)
-	case "storage proof":
+	case "storageProof":
 		res.Resolution = new(V2StorageProof)
 	case "finalization":
 		res.Resolution = new(V2FileContractFinalization)


n8maninger (approved):

n8maninger:
It would be nice to have unexported consts for these instead of magic strings. If you want to include that in this PR, it’d be appreciated. Otherwise, I’ll tackle it in a follow up.

lukechampine:
One way I've done that in the past is with a method on each of the enum types, i.e.

```go
type V2FileContractResolutionType interface {
	isV2FileContractResolution()
	fieldName() string
}
func (*V2FileContractFinalization) fieldName() string { return "finalization" }
func (*V2FileContractRenewal) fieldName() string      { return "renewal" }
func (*V2StorageProof) fieldName() string             { return "storageProof" }
func (*V2FileContractExpiration)fieldName() string    { return "expiration" }
```
Then you can do:
```go
func (res V2FileContractResolution) MarshalJSON() ([]byte, error) {
	return json.Marshal(struct {
		Parent     V2FileContractElement        `json:"parent"`
		Type       string                       `json:"type"`
		Resolution V2FileContractResolutionType `json:"resolution"`
	}{res.Parent, res.Resolution.fieldName(), res.Resolution})
}
```

The `UnmarshalJSON` side is kinda ugly, though:
```go
func (res *V2FileContractResolution) UnmarshalJSON(b []byte) error {
	var p struct {
		Parent     V2FileContractElement
		Type       string
		Resolution json.RawMessage
	}
	if err := json.Unmarshal(b, &p); err != nil {
		return err
	}
	switch p.Type {
	case (*V2FileContractRenewal)(nil).fieldName():
		res.Resolution = new(V2FileContractRenewal)
	case (*V2StorageProof)(nil).fieldName():
		res.Resolution = new(V2StorageProof)
	case (*V2FileContractFinalization)(nil).fieldName():
		res.Resolution = new(V2FileContractFinalization)
	case (*V2FileContractExpiration)(nil).fieldName():
		res.Resolution = new(V2FileContractExpiration)
	default:
		return fmt.Errorf("unknown file contract resolution type %q", p.Type)
	}
	if err := json.Unmarshal(p.Resolution, res.Resolution); err != nil {
		return err
	}
	res.Parent = p.Parent
	return nil
}
```
I don't like having to define extra methods, either. So overall I'm ambivalent as to whether this is better or worse than hardcoding the strings. ‾\\\_(ツ)\_/‾

PR #184, opened 2024-07-19
types: Omit payout field from FileContractRevision JSON

lukechampine:
See https://github.com/SiaFoundation/core-rs/pull/33

### DIFF
--- a/types/types.go
+++ b/types/types.go
@@ -1034,6 +1034,45 @@ func (sig Signature) MarshalText() ([]byte, error) { return marshalHex("sig", si
 func (sig *Signature) UnmarshalText(b []byte) error { return unmarshalHex(sig[:], "sig", b) }
 
 // MarshalJSON implements json.Marshaler.
+func (fcr FileContractRevision) MarshalJSON() ([]byte, error) {
+	return json.Marshal(struct {
+		ParentID         FileContractID   `json:"parentID"`
+		UnlockConditions UnlockConditions `json:"unlockConditions"`
+		Filesize         uint64           `json:"filesize"`
+		FileMerkleRoot   Hash256          `json:"fileMerkleRoot"`
+		WindowStart      uint64           `json:"windowStart"`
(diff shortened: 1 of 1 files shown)


types/encoding.go
```diff
@@ -1288,8 +1288,8 @@ func (ren *V2FileContractRenewal) DecodeFrom(d *Decoder) {
 }
 
 // DecodeFrom implements types.DecoderFrom.
-func (fcf V2FileContractFinalization) DecodeFrom(d *Decoder) {
-	(*V2FileContract)(&fcf).DecodeFrom(d)
+func (fcf *V2FileContractFinalization) DecodeFrom(d *Decoder) {
+	(*V2FileContract)(fcf).DecodeFrom(d)
```
lukechampine, line 1292: This fixes the bug found here https://github.com/SiaFoundation/walletd/pull/154/files#r1676487216

types/types.go
```diff
@@ -1033,6 +1033,45 @@ func (sig Signature) MarshalText() ([]byte, error) { return marshalHex("sig", si
 // UnmarshalText implements encoding.TextUnmarshaler.
 func (sig *Signature) UnmarshalText(b []byte) error { return unmarshalHex(sig[:], "sig", b) }
 
+// MarshalJSON implements json.Marshaler.
+func (fcr FileContractRevision) MarshalJSON() ([]byte, error) {
```
n8maninger, line 1037: This feels like it’d be better to separate the types over the doc string that mentions the sentinel value and custom Marshal methods to ignore the field

lukechampine: I don't follow...?

n8maninger: Apparently I’m just crazy’

lukechampine: 😆 

It *is* slightly gross to embed a type and special-case one of its fields, but arguably this is just a mistake in Sia v1 -- revisions *should* embed the entire contract. A consensus rule can enforce that the payout value doesn't change. v2 fixes this :tm:

PR #187, opened 2024-08-07
consensus: Add (*MidState).forEachRevertedElementLeaf

lukechampine:
When iterating over the elements in a `MidState`, we need to be careful with regard to contract revisions. When *applying* a block, we want to report the latest revision; but when *reverting*, we want to report the *earliest*. Unfortunately, this asymmetry was overlooked, leading to a rather pernicious bug: when a block containing a revision is reverted, the Merkle tree is still contains the revision, not the original contract.

It's pernicious because A) reorgs are fairly uncommon, and B) the reorg has to contain a revision, and also C) redundancy in the DB code means that the invalid hashes may be overwritten by valid ones, masking the problem. That's likely why we only discovered this bug after adding a sanity check that validates every block supplement. Sure, our consensus tests *are* a bit lacking when it comes to file contracts -- I don't think there are *any* involving both a revision and a reorg! -- but the nature of this bug makes me wonder if it would have eluded detection anyway.

If you've been running a `core` node long enough to observe a reorg, there's a decent chance your Merkle tree is (very slightly) wrong. This can *probably* be fixed by updating and then manually reverting and re-applying a lot of blocks... but resyncing from genesis is the best way to be sure your state is ok. 🙃

### DIFF
--- a/consensus/merkle.go
+++ b/consensus/merkle.go
@@ -360,8 +360,7 @@ func (acc *ElementAccumulator) applyBlock(updated, added []elementLeaf) (eau ele
 
 // revertBlock modifies the proofs of supplied elements such that they validate
 // under acc, which must be the accumulator prior to the application of those
-// elements. All of the elements will be marked unspent. The accumulator itself
-// is not modified.
+// elements. The accumulator itself is not modified.
 func (acc *ElementAccumulator) revertBlock(updated, added []elementLeaf) (eru elementRevertUpdate) {
 	eru.updated = updateLeaves(updated)
 	eru.numLeaves = acc.NumLeaves
--- a/consensus/state.go
+++ b/consensus/state.go
@@ -721,7 +721,7 @@ func (ts V1TransactionSupplement) siacoinElement(id types.SiacoinOutputID) (sce 
 	return
 }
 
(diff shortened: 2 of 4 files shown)


n8maninger:
Should add a test creating, applying and reverting all the different elements and validating the accumulator state after each

ChrisSchinnerl (commented):
I wonder if there is a nice way to validate the accumulator after every update in test builds. To get that implicit sanity check both here as well as part of `renterd` and `hostd` integration tests.

lukechampine:
> Should add a test with create -> apply -> revert -> apply for all the different elements and validate the accumulator state after each

Sounds like @chris124567 will be writing helpers for this shortly.

> I wonder if there is a nice way to validate the accumulator after every update in test builds.

You could do something like this, at the bottom of `consensus.ApplyBlock` and `consensus.RevertBlock`:
```go
for _, sce := range ms.sces {
	if !s.Elements.containsLeaf(siacoinLeaf(&sce, ms.isSpent(sce.ID))) {
		panic("consensus: siacoin element not found in accumulator after apply")
	}
}
for _, sfe := range ms.sfes {
	// ... etc ...
```
As you know, I am deathly allergic to build tags, but maybe we could stick a flag in `consensus.Network`? :thinking:

In any case, I'm in favor of merging this fix now, and addressing the rest in a follow-up.

PR #188, opened 2024-08-11
consensus: Disallow contract revision after proof window opens

lukechampine:
Slightly worrisome that I missed this check when porting `siad`... 👀
This demonstrates why it's insufficient to rely on the fact that `core` can sync the existing chain: those are all *valid* blocks. Syncing mainnet tells us nothing about whether `core` will behave the same as `siad` with respect to *invalid* blocks. So we should probably use the fuzzer to pass blocks to both `core` and `siad`, checking that either they both accept or both reject.


n8maninger:
Can you add a regression test with some examples of good and bad blocks?

lukechampine:
done

n8maninger (approved):

PR #196, opened 2024-09-05
Add revision ordering test

n8maninger:


lukechampine:
good find. I'm not really sure how to handle this case. There are a few options:
1. Disallow it entirely, i.e. enforce that a contract is revised at most once per block. This is obviously inconvenient.
1. Track the revisions within a block, and require the `Parent` field to reflect the most recent revision within a block. This is a bit awkward because the `Parent` won't actually be present in the accumulator.
1. Leave the `Parent` field as-is, but track the revisions so that we can reject lower revision numbers. This feels a bit contradictory: on one hand, since the `Parent` of all revisions is the same, the implication is that only one of those is the "real" revision. On the other hand, if only one revision is "real," why do we care if a later revision has a lower revision number, since it won't be getting applied anyway?
1. Mark as `wontfix` -- you could make the case that, as long as the revision number increases at block-granularity, it doesn't matter if it decreases at transaction-granularity. This is weird because it lets you create multiple revisions of a contract that have the same revision number.

I'm leaning towards 2; I like that it forces the transaction creator to acknowledge the prior revision. And there's some precedent: It's similar to the check we have for ephemeral siacoin outputs (which are also not in the accumulator). Thoughts?

n8maninger:
I think option one would be okay. If a renter and host both broadcast different versions of a contract, the higher one should always be prefered. The tpool would have to be modified to eject lower revisions though (maybe it should anyway?). Three would be my second choice. It doesn't restrict revisions too much and doesn't add an extra step to broadcasting a revision of checking the tpool for an ephemeral parent.

> On the other hand, if only one revision is "real," why do we care if a later revision has a lower revision number, since it won't be getting applied anyway

I might be misunderstanding this comment, but it looks like the lower revision *is* getting applied to the consensus state. If the second lower revision was ignored, I think we could safely mark this as wontfix. As it stands, you can overwrite a higher revision number with a lower one as long as the lower one is in a later transaction in the same block.

lukechampine:
Yeah, I think I agree. Although option 1 is probably ok in practice, it feels unnecessarily restrictive; of the others, option 3 is the most convenient, and it doesn't have any practical drawbacks, only conceptual ones.

PR #199, opened 2024-09-10
types: Move ID output of StateElement

lukechampine:
This is an experimental change that gives each element type an ID specific to that type, rather than all elements using `types.Hash256` as their ID type. That is, `sce.ID` is now a `types.SiacoinOutputID`. This prevents awkward casts in a few places, namely when deriving output IDs like `types.SiafundOutputID(sfi.Parent.ID).V2ClaimOutputID())` -- we can now write `sfi.Parent.ID.V2ClaimOutputID()` instead.

The second commit un-embeds the `StateElement` field from each element type. Embedding is best avoided unless there are significant benefits, and in this case the only significant benefit was convenient access to the `ID` field, which is now moot.

The combined effect on JSON encodings is:
```json
old:
{
    "id": "h:9373b2995a6f5d47d63aaf22dbbfe92aa43b84551fe5de1e08341fd02dd100ec",
    "leafIndex": 12,
    "merkleProof": [
        "h:ee261e985aefd3c1f849310dc697b3b4c5d9227b48507d7d4399adefc90708ad",
        "h:f00470b20ca5da851e86ba6793bd967eac56bb7501b1189336b31384c7652d4a"
    ]
    "siacoinOutput": {
        "address": "111a37d4aedbbf671e6ffea289e1bcb8521ab9433fd2e26bce1432696943cbf26612a970021b",
        "value": 10000000000000000000
    },
    "maturityHeight": 0
}
new:
{
    "stateElement": {
        "leafIndex": 12,
        "merkleProof": [
            "h:ee261e985aefd3c1f849310dc697b3b4c5d9227b48507d7d4399adefc90708ad",
            "h:f00470b20ca5da851e86ba6793bd967eac56bb7501b1189336b31384c7652d4a"
        ]
    },
    "id": "scoid:9373b2995a6f5d47d63aaf22dbbfe92aa43b84551fe5de1e08341fd02dd100ec",
    "siacoinOutput": {
        "address": "111a37d4aedbbf671e6ffea289e1bcb8521ab9433fd2e26bce1432696943cbf26612a970021b",
        "value": 10000000000000000000
    },
    "maturityHeight": 0
}
```
This definitely breaks compatibility, so we'd have to weigh whether it's worth the annoyance.

### DIFF
--- a/consensus/state.go
+++ b/consensus/state.go
@@ -589,12 +589,12 @@ func (s State) AttestationSigHash(a types.Attestation) types.Hash256 {
 // A MidState represents the state of the chain within a block.
 type MidState struct {
 	base               State
-	created            map[types.Hash256]int // indices into element slices
-	spends             map[types.Hash256]types.TransactionID
-	revs               map[types.Hash256]*types.FileContractElement
-	res                map[types.Hash256]bool
-	v2revs             map[types.Hash256]*types.V2FileContractElement
-	v2res              map[types.Hash256]types.V2FileContractResolutionType
+	created            map[types.ElementID]int // indices into element slices
+	spends             map[types.ElementID]types.TransactionID
+	revs               map[types.FileContractID]*types.FileContractElement
+	res                map[types.FileContractID]bool
+	v2revs             map[types.FileContractID]*types.V2FileContractElement
+	v2res              map[types.FileContractID]types.V2FileContractResolutionType
 	siafundPool        types.Currency
 	foundationPrimary  types.Address
 	foundationFailsafe types.Address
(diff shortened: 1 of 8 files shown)


n8maninger:
This looks like a job for generics! /s
```
// A StateElement is a generic element within the state accumulator.
type StateElement[T ~[32]byte] struct {
	ID          T         `json:"id"` // SiacoinOutputID, FileContractID, etc.
	LeafIndex   uint64    `json:"leafIndex"`
	MerkleProof []Hash256 `json:"merkleProof,omitempty"`
}
```

lukechampine:
does the encoding still work? Because honestly I think I'd be ok with that

lukechampine:
Update: the encoding does work, but after fixing up 90% of the code, I hit this...
```go
func (au ApplyUpdate) UpdateElementProof(e *types.StateElement) {
	au.eau.updateElementProof(e)
}
```
Go doesn't support generic methods, so we'd have to define a top-level function (`ApplyElementProof`?) for this, which sucks for various reasons (e.g. you can't define a `ProofUpdater` interface).

After reflecting on this some more, I think the approach taken in this PR is correct: `ID` should not be a field of `StateElement`. The ID itself actually has no bearing on the state accumulator whatsoever, except for its contribution to the leaf hash -- which is parameterized by type anyway.

PR #200, opened 2024-09-13
RHP4

n8maninger:


consensus/validation.go
```diff
@@ -760,6 +760,8 @@ func validateV2FileContracts(ms *MidState, txn types.V2Transaction) error {
 			return fmt.Errorf("does not increase revision number (%v -> %v)", cur.RevisionNumber, rev.RevisionNumber)
 		case !revOutputSum.Equals(curOutputSum):
 			return fmt.Errorf("modifies output sum (%d H -> %d H)", curOutputSum, revOutputSum)
+		case rev.MissedHostValue.Cmp(cur.MissedHostValue) > 0:
+			return fmt.Errorf("has missed host value (%d H) exceeding old value (%d H)", rev.MissedHostValue, cur.MissedHostValue)
```
lukechampine, line 764: good catch. MissedHost should be free to decrease (meaning: the host has put up more collateral), but not increase -- the host shouldn't be able to rescind the collateral they put up.

rhp/v4/merkle.go
```diff
@@ -0,0 +1,190 @@
+package rhp
```
lukechampine, line 1: could this file just re-exported the stuff in rhp/v2?

rhp/v4/validation.go
```diff
@@ -0,0 +1,127 @@
...
+)
+
+// Validate validates a read sector request.
+func (req *RPCReadSectorRequest) Validate() error {
+	switch {
+	case req.Length == 0:
+		return errors.New("length must be greater than 0")
+	case req.Offset+req.Length > SectorSize:
+		return errors.New("read request exceeds sector bounds")
+	case (req.Offset+req.Length)%LeafSize != 0:
```
lukechampine, line 17: did we consider making `Offset` and `Length` multiples of `LeafSize`? (i.e. rename them to `LeafIndex` and `NumLeaves`?)

rhp/v4/rhp.go
```diff
@@ -0,0 +1,640 @@
...
+		Token      AccountToken `json:"token"`
+		Duration   uint64       `json:"duration"`
+		DataLength uint64       `json:"dataLength"` // extended to SectorSize by host
+	}
+
+	// RPCWriteSectorRequest implements Request.
+	RPCWriteSectorRequest struct {
+		Prices   HostPrices   `json:"prices"`
+		Token    AccountToken `json:"token"`
+		Duration uint64       `json:"duration"`
```
lukechampine, line 361: what is this field for again?

rhp/v4/rhp.go
```diff
@@ -0,0 +1,640 @@
...
+		cost = req.Prices.StoragePrice.Mul64(size * duration)
+		collateral = req.Prices.Collateral.Mul64(size * duration)
+	}
+
+	// pay with contract is always called because it increments the revision number
+	err := PayWithContract(&fc, cost, collateral)
+	if err != nil {
+		return fc, err
+	}
+	fc.FileMerkleRoot = resp.Proof[len(resp.Proof)-1] // TODO get merkle root
```
lukechampine, line 554: 🫡

rhp/v4/rhp.go
```diff
@@ -0,0 +1,640 @@
...
+func ReviseForFundAccount(fc types.V2FileContract, amount types.Currency) (types.V2FileContract, error) {
+	err := PayWithContract(&fc, amount, types.ZeroCurrency)
+	return fc, err
+}
+
+// MinRenterAllowance returns the minimum allowance required to justify the given
+// host collateral.
+func MinRenterAllowance(hp HostPrices, duration uint64, collateral types.Currency) types.Currency {
+	maxCollateralBytes := collateral.Div(hp.Collateral).Div64(duration)
+	return hp.StoragePrice.Mul64(duration).Mul(maxCollateralBytes).Mul64(9).Div64(10) // 10% buffer
```
lukechampine, line 574: I was hoping to eliminate all buffers in v4 🤔

rhp/v4/rhp.go
```diff
@@ -0,0 +1,640 @@
...
+
+// GenerateAccount generates a pair of private key and Account from a secure
+// entropy source.
+func GenerateAccount() (types.PrivateKey, Account) {
+	sk := types.GeneratePrivateKey()
+	return sk, Account(sk.PublicKey())
+}
+
+type (
+	// RPCSettingsRequest implements Request.
```
lukechampine, line 220: I think we only have `Object` now, not `Request`

lukechampine (commented):
looks good. Are there any big TODOs other than calculating the new Merkle root in `ReviseForModifySectors`?

rhp/v4/merkle.go
n8maninger: Probably a lot of it

rhp/v4/rhp.go
n8maninger: Sectors are written to disk without being associated with a contract. This gives the renter control over when the "temporary" sector expires

rhp/v4/encoding.go
```diff
@@ -0,0 +1,699 @@
...
+			flags |= uint64(1) << (i % 64)
+		}
+		if (i+1)%64 == 0 {
+			e.WriteUint64(flags)
+			flags = 0
+		}
+	}
+	if len(r.Accepted)%64 != 0 {
+		e.WriteUint64(flags)
+	}
```
lukechampine, line 469: fun, but surely overkill; if we use the standard 1-byte-per-bool encoding, a 1000-sector append still easily fits in one network packet.

rhp/v4/rhp.go
```diff
@@ -0,0 +1,754 @@
...
+	B    uint64        `json:"b,omitempty"`    // Swap
+	N    uint64        `json:"n,omitempty"`    // Trim
+}
+
+// WriteAction types.
+const (
+	ActionSwap = iota + 1
+	ActionTrim
+	ActionUpdate
+)
```
lukechampine, line 167: IIRC you mentioned something about RPCModifySectors supporting fewer action types? Need to know for the Merkle proof code.

n8maninger: RPC Modify supports swap, trim, and update. Append is now a standalone RPC. The benefit being that the renter may not be 100% sure a sector is stored on the host when pinning roots to a contract. The renter can batch a large number of roots without concern that it will fail in the middle and have to retry.

rhp/v4/encoding.go
n8maninger: Fun removed

rhp/v4/rhp.go
```diff
@@ -0,0 +1,754 @@
...
+}
+
+// A ModifyAction adds or modifies sectors within a contract.
+type ModifyAction struct {
+	Type uint8         `json:"type"`
+	Root types.Hash256 `json:"root,omitempty"` // Update
+	A    uint64        `json:"a,omitempty"`    // Swap
+	B    uint64        `json:"b,omitempty"`    // Swap
+	N    uint64        `json:"n,omitempty"`    // Trim, Update
+}
```
lukechampine, line 160: How does `ActionUpdate` work? From the server code in `coreutils`, it looks like it overwrites the sector root at index `A` with `Root`. That's very different from the old `RPCWriteActionUpdate`, which overwrote sector *data*. Also, if `Root` is the root of a sector that the host isn't storing, the host won't be able to complete a storage proof for it!

I think we should consider scrapping the general-purpose `ModifyAction` design and instead hard-code the specific actions we know we want. When RHPv2 was designed, we envisioned something more like a distributed filesystem: sector 0 was used to store metadata about the other sectors, and diff proofs allowed you to atomically update both the data and the metadata. Nowadays (AFAICT) we only do two things: append new sectors, and delete old sectors (via swap + trim). Append is already a separate RPC now; perhaps instead of a generic `RPCModify`, we only need `RPCSwapAndTrim`. That would simplify the diff proofs a bit too.

More radically, we could get rid of trimming. Instead, "deleted" sectors are overwritten with zeroes, and when you upload new sectors, they replace zeroed sectors if available. This matches how the payment works anyway: renters don't get refunded when they trim. I think we discussed this option at the RHP summit, and presumably we rejected it, but I don't remember the specific objections.

n8maninger: The presumption is that the sector will always exist on the host already through RPC write. Similar to append, the server should be checking if the host has a sector before allowing the update.

Update is currently necessary for cost efficient deletion. After trimming, there isn’t a way to track the max size of the contract so future appends will cost additional SC and we can’t do the atomic append+swap+trim any more. It does feel out of place in the new rpc design.

rhp/v4/merkle.go
```diff
@@ -106,42 +106,30 @@ func VerifySectorRootsProof(proof, sectorRoots []types.Hash256, numSectors, star
...
-				A:    a.N,
-			}
-		case ActionUpdate:
-			rhp2Actions[i] = rhp2.RPCWriteAction{
-				Type: rhp2.RPCWriteActionUpdate,
-				A:    a.N,
-				Data: nil, // TODO
-			}
-		}
+func convertFreeActions(freed []uint64, numSectors uint64) []rhp2.RPCWriteAction {
```
n8maninger, line 109: This is going to need to be rewritten after we delete RHP2 anyway. Can this be inlined and get rid of the generic rhp2.RPCWriteAction for good?

rhp/v4/rhp.go
```diff
@@ -0,0 +1,768 @@
...
+	"fmt"
+	"io"
+	"time"
+
+	"go.sia.tech/core/consensus"
+	"go.sia.tech/core/types"
+)
+
+const (
+	proofWindow = 144 // 24 hours
```
peterjan, line 15: Export it?

rhp/v4/rhp.go
```diff
@@ -0,0 +1,768 @@
...
+	TotalStorage        uint64         `json:"totalStorage"`
+	Prices              HostPrices     `json:"prices"`
+}
+
+// An Account represents an ephemeral balance that can be funded via contract
+// revision and spent to pay for RPCs.
+type Account types.PublicKey
+
+// String implements fmt.Stringer.
+func (a Account) String() string { return fmt.Sprintf("acct:%x", a[:]) }
```
peterjan, line 190: (TIL `acct` is the more formal and widely recognised abbreviation of account)

rhp/v4/validation.go
```diff
@@ -0,0 +1,222 @@
...
+	switch {
+	case req.Duration == 0:
+		return errors.New("duration must be greater than 0")
+	case req.DataLength == 0:
+		return errors.New("sector must not be empty")
+	case req.DataLength%LeafSize != 0:
+		return errors.New("sector must be segment aligned")
+	case req.DataLength > SectorSize:
+		return errors.New("sector exceeds sector bounds")
+	case req.Duration > maxDuration:
```
peterjan, line 44: `nit` group cases by field

rhp/v4/merkle.go
```diff
@@ -0,0 +1,135 @@
...
+func BuildSectorProof(sector *[SectorSize]byte, start, end uint64) []types.Hash256 {
+	return rhp2.BuildProof(sector, start, end, nil)
+}
+
+// A RangeProofVerifier allows range proofs to be verified in streaming fashion.
+type RangeProofVerifier = rhp2.RangeProofVerifier
+
+// NewRangeProofVerifier returns a RangeProofVerifier for the sector range
+// [start, end).
+func NewRangeProofVerifier(start, end uint64) *RangeProofVerifier {
```
peterjan, line 51: Should we re-export the `RangeProofVerifier` exactly as is or rename it, similar to how we're more or less hiding the term "diff proof" now, or better choose not to export the concept of a diff proof.

peterjan (commented):
LGTM 👍

rhp/v4/merkle.go
ChrisSchinnerl: Probably fine to keep this but I don't feel strongly about either approach

lukechampine: I intend to go HAM on the Merkle stuff once the dust has fully settled

rhp/v4/rhp.go
```diff
@@ -0,0 +1,768 @@
...
+
+	"go.sia.tech/core/consensus"
+	"go.sia.tech/core/types"
+)
+
+const (
+	proofWindow = 144 // 24 hours
+
+	// SectorSize is the size of one sector in bytes.
+	SectorSize = 1 << 22 // 4 MiB
```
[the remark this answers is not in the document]
lukechampine: yeah I'm a little torn too. On one hand, I like things being clean and self-contained. On the other hand, "forwarding" these constants/types/functions from `rhp2` makes it more obvious which aspects of the RHP have changed and which haven't. I think my recommendation is to reuse `rhp2` initially -- then it'll at least be in the git history. We can inline everything once we formally deprecate RHPv2.

rhp/v4/rhp.go
```diff
@@ -0,0 +1,768 @@
...
+	renewal.FinalRevision.RenterSignature = types.Signature{}
+	renewal.FinalRevision.HostSignature = types.Signature{}
+
+	// create the new contract
+	renewal.NewContract = fc
+	renewal.NewContract.RevisionNumber = 0
+	renewal.NewContract.RenterSignature = types.Signature{}
+	renewal.NewContract.HostSignature = types.Signature{}
+	// add the additional allowance and collateral
+	renewal.NewContract.RenterOutput.Value = fc.RenterOutput.Value.Add(rp.Allowance)
```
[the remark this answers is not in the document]
n8maninger: I'm not sure adding `Additional` is significantly better since a refreshed or renewed contract will both have at least `allowance` and `collateral` available to use.

ChrisSchinnerl (approved):

PR #208, opened 2024-09-26
Be explicit about subsidy existence

n8maninger:


lukechampine (approved):
lgtm, though probably the docstring should be updated as well. You could just delete the second line; the `exists` in the function signature implies that the output is only valid when it's `true`.

PR #209, opened 2024-10-03
Remove resolution txn restrictions

n8maninger:
Removes the need for a setup transaction for v2 contract resolutions

### DIFF
--- a/consensus/validation.go
+++ b/consensus/validation.go
@@ -679,16 +679,6 @@ func validateV2Siafunds(ms *MidState, txn types.V2Transaction) error {
 }
 
 func validateV2FileContracts(ms *MidState, txn types.V2Transaction) error {
-	// Contract resolutions are height-sensitive, and thus can be invalidated by
-	// shallow reorgs; to minimize disruption, we require that transactions
-	// containing a resolution do not create new outputs. Creating, revising or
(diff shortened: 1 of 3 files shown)


lukechampine:
Why *did* we need a setup transaction? Is it purely to pay the miner fee? If so, perhaps some of the funds in a contract resolution could be used to pay the fee...? 🤔

n8maninger:
Renter allowance, host collateral, and miner fee.

n8maninger:
Looks like this would also have a side-effect of enabling the host/renter rollover to "escape"...

lukechampine:
Another option is to timelock the outputs, i.e. the change output from a resolution wouldn't be spendable for 144 blocks. That might be confusing from a UX perspective though.

n8maninger:
@lukechampine are you okay with this as is?

lukechampine:
yeah, I think so, as long as we're a bit more cautious about immediately chaining those outputs into new transactions.

lukechampine:
c.f. https://github.com/SiaFoundation/core/pull/43...
I'm thinking that we should probably not allow contract outputs to be ephemeral. That's enough of a deterrent to prevent the most egregious forms of output chaining.

PR #211, opened 2024-10-07
types,consensus: Make contract finalization unilateral

lukechampine:
Previously, a `V2FileContractFinalization` was a final revision of a contract, mutually agreed upon by the renter and host. However, there are scenarios in which a renter may want to unilaterally end a contract, such as when they want to leave the network entirely. This PR changes `V2FileContractFinalization` to be a signature: specifically, the renter's signature of the current revision, with its revision number set to `MaxRevisionNumber`. This addresses multiple concerns:

- Clearly the renter must not be allowed to finalize a contract revision that lacks a valid host signature. The approach in this PR forces the renter to finalize the *existing* state of the contract, so the host is not at risk.
- The host must not be allowed to finalize contracts at all, since they would be incentivized to do so as soon as they detect data loss (to avoid forfeiting their collateral). Requiring the signature to use `MaxRevisionNumber` prevents the host from replaying a previous renter signature.
- Mutual finalization remains possible: the renter and host simply agree to submit a revision, and then the renter submits the finalization. (Currently, it's possible to do both within the same block, though not the same transaction.)

### DIFF
--- a/consensus/update.go
+++ b/consensus/update.go
@@ -547,7 +547,7 @@ func (ms *MidState) ApplyV2Transaction(txn types.V2Transaction) {
 		case *types.V2StorageProof:
 			renter, host = fc.RenterOutput, fc.HostOutput
 		case *types.V2FileContractFinalization:
-			renter, host = r.RenterOutput, r.HostOutput
+			renter, host = fc.RenterOutput, fc.HostOutput
 		case *types.V2FileContractExpiration:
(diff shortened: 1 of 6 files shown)


lukechampine:
Note that finalization [was originally added](https://github.com/SiaFoundation/core/pull/31) as a way to implement "renew and clear," back before we had an explicit `V2FileContractRenewal` type.

n8maninger (approved):

PR #212, opened 2024-10-07
RPC refresh

n8maninger:
Add explicit refresh RPC for more efficient and correct contract refreshes

### DIFF
--- a/rhp/v4/rhp.go
+++ b/rhp/v4/rhp.go
@@ -266,8 +266,8 @@ type (
 		TransactionSet []types.V2Transaction `json:"transactionSet"`
 	}
 
-	// RPCRefreshContractParams includes the contract details required to create
-	// a renewal.
+	// RPCRefreshContractParams includes the contract details required to refresh
+	// a contract.
 	RPCRefreshContractParams struct {
 		ContractID types.FileContractID `json:"contractID"`
 		Allowance  types.Currency       `json:"allowance"`
@@ -733,8 +733,6 @@ func RefreshContract(fc types.V2FileContract, prices HostPrices, rp RPCRefreshCo
 	// clear the old contract
 	renewal.FinalRevision = fc
 	renewal.FinalRevision.RevisionNumber = types.MaxRevisionNumber
-	renewal.FinalRevision.Filesize = 0
-	renewal.FinalRevision.FileMerkleRoot = types.Hash256{}
 	renewal.FinalRevision.RenterSignature = types.Signature{}
 	renewal.FinalRevision.HostSignature = types.Signature{}
 
--- a/rhp/v4/rhp.go
+++ b/rhp/v4/rhp.go
@@ -20,18 +20,19 @@ const (
 
 // RPC identifiers.
 var (
-	RPCAccountBalanceID = types.NewSpecifier("AccountBalance")
-	RPCFormContractID   = types.NewSpecifier("FormContract")
-	RPCFundAccountsID   = types.NewSpecifier("FundAccounts")
-	RPCLatestRevisionID = types.NewSpecifier("LatestRevision")
-	RPCAppendSectorsID  = types.NewSpecifier("AppendSectors")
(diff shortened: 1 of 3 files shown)


rhp/v4/rhp.go
```diff
@@ -677,3 +727,30 @@ func NewRenewal(fc types.V2FileContract, prices HostPrices, rp RPCRenewContractP
...
+
+// RefreshContract creates a new contract renewal from an existing contract revision
+func RefreshContract(fc types.V2FileContract, prices HostPrices, rp RPCRefreshContractParams) (renewal types.V2FileContractRenewal) {
+	// clear the old contract
+	renewal.FinalRevision = fc
+	renewal.FinalRevision.RevisionNumber = types.MaxRevisionNumber
+	renewal.FinalRevision.Filesize = 0
+	renewal.FinalRevision.FileMerkleRoot = types.Hash256{}
+	renewal.FinalRevision.RenterSignature = types.Signature{}
+	renewal.FinalRevision.HostSignature = types.Signature{}
```
lukechampine, line 739: I don't think it's necessary to clear these fields; consensus should treat it as a valid resolution regardless.

rhp/v4/rhp.go
```diff
@@ -265,6 +266,39 @@ type (
 		TransactionSet []types.V2Transaction `json:"transactionSet"`
 	}
 
+	// RPCRefreshContractParams includes the contract details required to create
```
peterjan, line 269: `"required to refresh a contract."`?

rhp/v4/validation.go
```diff
@@ -149,6 +151,38 @@ func (req *RPCRenewContractRequest) Validate(pk types.PublicKey, tip types.Chain
 	}
 }
 
+// Validate validates a renew contract request. Prices are not validated
```
peterjan, line 154:
```suggestion
// Validate validates a refresh contract request. Prices are not validated
```

rhp/v4/validation.go
```diff
@@ -149,6 +151,38 @@ func (req *RPCRenewContractRequest) Validate(pk types.PublicKey, tip types.Chain
...
+	// host's locked collateral and the contract duration
+	minRenterAllowance := MinRenterAllowance(hp, expirationHeight-req.Prices.TipHeight, req.Refresh.Collateral)
+
+	switch {
+	case req.Refresh.Allowance.IsZero():
+		return errors.New("allowance is zero")
+	case req.Refresh.Collateral.Cmp(maxCollateral) > 0:
+		return fmt.Errorf("collateral %v exceeds max collateral %v", req.Refresh.Collateral, maxCollateral)
+	case req.Refresh.Allowance.Cmp(minRenterAllowance) < 0:
+		return fmt.Errorf("allowance %v is less than minimum %v for collateral", req.Refresh.Allowance, minRenterAllowance)
```
peterjan, line 180: I'm guessing "for collateral" is a typo in the error message and it's meant to be "allowance x is less than min allowance y". It's repeated three times in the repo.

rhp/v4/validation.go
```diff
@@ -149,6 +151,38 @@ func (req *RPCRenewContractRequest) Validate(pk types.PublicKey, tip types.Chain
...
+
+	// validate the contract fields
+	hp := req.Prices
+	// calculate the minimum allowance required for the contract based on the
+	// host's locked collateral and the contract duration
+	minRenterAllowance := MinRenterAllowance(hp, expirationHeight-req.Prices.TipHeight, req.Refresh.Collateral)
+
+	switch {
+	case req.Refresh.Allowance.IsZero():
+		return errors.New("allowance is zero")
```
peterjan, line 176: We have two styles of error messages. Sometimes we state the value and indicate it's invalid, while other times we specify that 'value X must adhere to condition Y,' and explain that the condition was not met. I prefer the latter, where we would error out here with "allowance must be greater than zero". I'm going to get shamed for this `nit` but I don't care.

lukechampine (approved):

PR #219, opened 2024-10-21
RHP4 spec changes

### LINKED ISSUE #218: Update RHP4 RPC types
Update the RHP4 RPC types to support the most recent changes to the spec.

- [ ] Add RPC delete types
- [ ] Remove RPC modify types

n8maninger:
Fixes #218 

Requires #220

### DIFF
--- a/consensus/validation.go
+++ b/consensus/validation.go
@@ -715,6 +715,8 @@ func validateV2FileContracts(ms *MidState, txn types.V2Transaction) error {
 
 	validateContract := func(fc types.V2FileContract, renewal bool) error {
 		switch {
+		case fc.Filesize > fc.Capacity:
+			return fmt.Errorf("has filesize (%v) exceeding capacity (%v)", fc.Filesize, fc.Capacity)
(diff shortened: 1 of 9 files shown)


consensus/validation.go
```diff
@@ -715,6 +715,8 @@ func validateV2FileContracts(ms *MidState, txn types.V2Transaction) error {
 
 	validateContract := func(fc types.V2FileContract, renewal bool) error {
 		switch {
+		case fc.Filesize > fc.Capacity:
```
ChrisSchinnerl, line 718: Could it be that we are missing `fc.Filesize % SectorSize == 0` checks? Same for capacity?

lukechampine: `SectorSize` is not a consensus concept, only an RHP one.

ChrisSchinnerl: @lukechampine what about `SegmentSize`?

ChrisSchinnerl (approved):

PR #220, opened 2024-10-21
types,consensus: Add Capacity field to V2FileContract

lukechampine:
This field clarifies some long-standing confusion around the economics of file contracts.

When you pay for storage in a file contract, you are buying that storage for the remainder of the contract duration. This means that, even if you delete some data, you do not get a refund. Why not? Well, mainly, how do you price the refund? If you use the host's current prices, the host can manipulate them in order to minimize the refund. On the other hand, if you use the original price, now you have to keep track of what the original price was -- and not just the price, but also *how long* each sector was originally stored for. That's a *lot* metadata, and both the renter *and* host need to track it accurately, or they'll disagree on the price and stop communicating entirely.

Okay, so we can't do refunds. But we also want a way for renters to get rid of data they don't care about. One way to do that is to *overwrite* the garbage data with data we *do* care about. For this to work, though, the renter has to keep track of the garbage sectors, which is annoying. In general, we want to minimize renter-side state (to facilitate light clients), so we shift responsibilities to the host and use cryptography to make sure the host doesn't cheat. So in RHPv4, that's the plan: have the *host* keep track of the garbage sectors, so that when you upload new data, the host can reuse existing capacity instead of increasing the contract size. This is why we need a capacity field: to ensure that the host doesn't lie about how much capacity is available.

### DIFF
--- a/consensus/validation.go
+++ b/consensus/validation.go
@@ -725,6 +725,8 @@ func validateV2FileContracts(ms *MidState, txn types.V2Transaction) error {
 
 	validateContract := func(fc types.V2FileContract, renewal bool) error {
 		switch {
+		case fc.Filesize > fc.Capacity:
+			return fmt.Errorf("has filesize (%v) exceeding capacity (%v)", fc.Filesize, fc.Capacity)
 		case fc.ProofHeight < ms.base.childHeight():
 			return fmt.Errorf("has proof height (%v) that has already passed", fc.ProofHeight)
 		case fc.ExpirationHeight <= fc.ProofHeight:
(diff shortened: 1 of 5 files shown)


n8maninger (approved):

lukechampine:
hmm... how do storage proofs interact with this, though? Shouldn't we use the capacity, not the filesize, to choose the proof index? But if so, the storage proof could be for an empty sector. Maybe that's okay? Feels a bit weird though.

consensus/validation.go
```diff
@@ -725,6 +725,8 @@ func validateV2FileContracts(ms *MidState, txn types.V2Transaction) error {
 
 	validateContract := func(fc types.V2FileContract, renewal bool) error {
 		switch {
+		case fc.Filesize > fc.Capacity:
+			return fmt.Errorf("has filesize (%v) exceeding capacity (%v)", fc.Filesize, fc.Capacity)
```
lukechampine, line 729: we could make this stricter -- `fc.Filesize != fc.Capacity`. There might be a reason to create a contract with extra capacity, though, idk.

n8maninger: I can't think of a good reason to restrict it and there might be a usecase for allowing it.

PR #224, opened 2024-10-29
Remove prefixes

n8maninger:
When working with these types the context is usually known. The prefixes make storing and comparing hex strings more difficult. Removes the prefixes from the all of the hex types except for `UnlockKey` and `PublicKey`.


lukechampine:
goodnight sweet prince. 😔 It was a promising idea, but in practice the costs outweighed the benefits.

lukechampine (approved):

PR #228, opened 2024-11-09
Refactor BlockHeader

lukechampine:
This unifies v1 and v2 block headers and promotes `BlockHeader` from `gateway` to `types`. It also changes how V2 block IDs are computed.

I had previously omitted the `ParentID` from the v2 block "header," because it is incorporated into the `Commitment` field (via the hash of the parent `consensus.State`). This left the first 32 bytes of the 80-byte hash preimage unused, so I filled them with a distinguisher string.

However, after reflecting on how headers are actually used in the Syncer, I think it's better to include the `ParentID` after all, even if it's slightly redundant and not *strictly* necessary. Here's my reasoning, in the form of a dialogue:

- 🤔 Foo: Imagine a peer sends you a v2 header, claiming that they have found the next block. You hash the header and verify that, indeed, it meets the proof-of-work target. You therefore stop mining the current block and begin mining the next one. Unbeknownst to you, this header does not attach to the old block! Now you are wasting hashrate until you detect this deception. If the header contained the `ParentID`, you could check that it matched your current block before abandoning it.
- 🤨 Bar: Including the `ParentID` doesn't prevent this class of attack, because the purported block could be invalid for other reasons that aren't detectable from the header alone. For example, it could have missing signatures. Moreover, this attack requires the attacker to expend a lot of hashrate; why wouldn't they use that hashrate to mine a valid block instead?
- 🤔 Foo: Aha, but that's where the `ParentID` is different! You *don't* need to expend a lot of hashrate mining a purposefully-invalid block; instead, you can *reuse* the header of a recently-mined block. That header will (probably) still have sufficient proof-of-work to meet the current target, so it will pass the initial validation check.
- 🤨 Bar: Okay, but one of the first checks you perform when receiving a relayed header is "have I seen this block before?" So if you try to cheat by reusing a recent block, it will simply be ignored.
- 🤔 Foo: ...fine, granted. But there's basically no downside to include the `ParentID` anyway. We weren't using that space anyway (but we need to preserve it, for miner compatibility), it's occasionally useful, and you don't need to twist your brain into knots convincing yourself that the system is still secure without it.
- 🤨 Bar: ...sure, whatever.

### DIFF
--- a/consensus/state.go
+++ b/consensus/state.go
@@ -609,7 +609,7 @@ type MidState struct {
 }
 
 func (ms *MidState) siacoinElement(ts V1TransactionSupplement, id types.SiacoinOutputID) (types.SiacoinElement, bool) {
-	if i, ok := ms.created[types.Hash256(id)]; ok {
+	if i, ok := ms.created[id]; ok {
 		return ms.sces[i], true
 	}
 	return ts.siacoinElement(id)
@@ -714,7 +714,7 @@ func (ts *V1TransactionSupplement) DecodeFrom(d *types.Decoder) {
 
 func (ts V1TransactionSupplement) siacoinElement(id types.SiacoinOutputID) (sce types.SiacoinElement, ok bool) {
 	for _, sce := range ts.SiacoinInputs {
-		if types.SiacoinOutputID(sce.ID) == id {
+		if sce.ID == id {
 			return sce, true
 		}
 	}
@@ -723,7 +723,7 @@ func (ts V1TransactionSupplement) siacoinElement(id types.SiacoinOutputID) (sce 
 
 func (ts V1TransactionSupplement) siafundElement(id types.SiafundOutputID) (sfe types.SiafundElement, ok bool) {
 	for _, sfe := range ts.SiafundInputs {
-		if types.SiafundOutputID(sfe.ID) == id {
+		if sfe.ID == id {
 			return sfe, true
 		}
 	}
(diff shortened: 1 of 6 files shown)


ChrisSchinnerl (commented):
I don't mind this change but we need an upgrade path that is not a panic due to the changed encoding. We can't stable release `renterd` (or any node software really) with the new core consensus if every update of the core dependency potentially causes a panic and requires manual user interaction.

lukechampine:
Agreed. Nothing here would trigger a panic, though; perhaps you're referring to https://github.com/SiaFoundation/coreutils/pull/116.

lukechampine:
I'm not sure I follow. Do you mean that this change breaks existing v2 testnets? Or that https://github.com/SiaFoundation/coreutils/pull/116 breaks existing `consensus.db`s?

n8maninger:
I think the primary concern is that a few other recent PRs broke existing `consensus.db` such that manual intervention is required over anything specific to this PR or https://github.com/SiaFoundation/coreutils/pull/116. I agree with your comment here: https://github.com/SiaFoundation/coreutils/pull/116/files#r1837380099 that these PR shouldn't cause a *new* issue. 

We do currently have a problem in the apps that use the coreutils chain db where we can't release an update without requiring manual intervention on the consensus database. That probably doesn't need to be tackled in either of these open PR, but it will need to be solved in a follow up at some point soon.

n8maninger (approved):

n8maninger:
It does look like we'll break Anagami since the V2BlockHeader encoding is changing, but I that's acceptable.

lukechampine:
Fair, I have been pretty cavalier about consensus changes lately. 😅 I'll try to rein it in as we approach release.

And yeah, changes to network stuff are less bothersome, since they'll resolve themselves once enough people update.

PR #230, opened 2024-11-11
JSON consistency

n8maninger:


types/types.go
```diff
@@ -1096,7 +1096,7 @@ func (fcr FileContractRevision) MarshalJSON() ([]byte, error) {
 		// Payout omitted; see FileContractRevision docstring
 		ValidProofOutputs  []SiacoinOutput `json:"validProofOutputs"`
 		MissedProofOutputs []SiacoinOutput `json:"missedProofOutputs"`
-		UnlockHash         Hash256         `json:"unlockHash"`
+		UnlockHash         Address         `json:"unlockHash"`
```
lukechampine, line 1099: IIRC I did this because this isn't "really" an `Address; sure, you can technically send coins to it, and the renter and host could spend those coins if they both sign the transaction, but for practical purposes these `UnlockConditions` are intended to be used for revising the `FileContract` and nothing else.

I'm fine with this change, though; no one really cares about the distinction, and it saves a cast in a few spots, so whatevs. In `V2FileContract` we ditch the arbitrary `UnlockConditions` in favor of hardcoded host and renter pubkeys, so it's a moot point there.

lukechampine (approved):

PR #231, opened 2024-11-13
Add ID field to Transaction and V2Transaction JSON encoding

n8maninger:
This adds the calculated ID for the transaction to the JSON marshaling of a transaction. This is primarily a convenience for API consumers that may or may not have the ability to calculate it easily. I believe `siad` added a calculate endpoint, but it's even more annoying to have to call 2 routes imo.

I decided not to override unmarshal and discard the ID for similar convenience, but it could be desirable to validate the deserialized transaction data against the provided ID. However, that would complicate broadcasting a transaction.

Most recent context, but this comes up quite a lot with integrators: https://discord.com/channels/809849352516141067/809858207064653894/1306241130408443915.

### DIFF
--- a/types/types.go
+++ b/types/types.go
@@ -404,7 +404,7 @@ type Transaction struct {
 //
 // For convenience, the transaction's ID is also calculated and included. This field is ignored during unmarshalling.
 func (txn Transaction) MarshalJSON() ([]byte, error) {
-	type jsonTxn Transaction
+	type jsonTxn Transaction // prevent recursion
 	return json.Marshal(struct {
 		ID TransactionID `json:"id"`
 		jsonTxn
@@ -704,7 +704,7 @@ type V2Transaction struct {
 //
 // For convenience, the transaction's ID is also calculated and included. This field is ignored during unmarshalling.
 func (txn V2Transaction) MarshalJSON() ([]byte, error) {
-	type jsonTxn V2Transaction
+	type jsonTxn V2Transaction // prevent recursion
 	return json.Marshal(struct {
 		ID TransactionID `json:"id"`
 		jsonTxn
--- a/types/types.go
+++ b/types/types.go
@@ -400,6 +400,38 @@ type Transaction struct {
 	Signatures            []TransactionSignature `json:"signatures,omitempty"`
 }
 
+// MarshalJSON implements json.Marshaller.
+//
+// json.Umarshaller is not implemented because the ID should be discarded.
+func (txn Transaction) MarshalJSON() ([]byte, error) {
+	jsonTxn := struct {
+		ID                    TransactionID          `json:"id"`
+		SiacoinInputs         []SiacoinInput         `json:"siacoinInputs,omitempty"`
+		SiacoinOutputs        []SiacoinOutput        `json:"siacoinOutputs,omitempty"`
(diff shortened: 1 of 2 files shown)


lukechampine:
Intriguing. In the past we've wrapped `Transaction` in API-specific types that add an `ID` field, but it's not hard to see the disadvantages of that approach.

Introducing "synthetic" fields like this to a marshalled object does make me a little uneasy; I don't like the fact that the encoding does not map 1-to-1 to the Go struct, and I don't like the unmarshal asymmetry either. But the benefits might outweigh the costs in this case.

types/types.go
```diff
@@ -400,6 +400,38 @@ type Transaction struct {
...
+		SiacoinInputs         []SiacoinInput         `json:"siacoinInputs,omitempty"`
+		SiacoinOutputs        []SiacoinOutput        `json:"siacoinOutputs,omitempty"`
+		FileContracts         []FileContract         `json:"fileContracts,omitempty"`
+		FileContractRevisions []FileContractRevision `json:"fileContractRevisions,omitempty"`
+		StorageProofs         []StorageProof         `json:"storageProofs,omitempty"`
+		SiafundInputs         []SiafundInput         `json:"siafundInputs,omitempty"`
+		SiafundOutputs        []SiafundOutput        `json:"siafundOutputs,omitempty"`
+		MinerFees             []Currency             `json:"minerFees,omitempty"`
+		ArbitraryData         [][]byte               `json:"arbitraryData,omitempty"`
+		Signatures            []TransactionSignature `json:"signatures,omitempty"`
```
lukechampine, line 418: finally, a legitimate usecase for embedding:
```go
type jsonTxn Transaction // prevent recursion
return json.Marshal(struct{
    ID TransactionID `json:"id"`
    jsonTxn
}{txn.ID(), jsonTxn(txn)})

File: types/types.go [hunk 2]
```diff
@@ -400,6 +400,38 @@ type Transaction struct {
 	Signatures            []TransactionSignature `json:"signatures,omitempty"`
 }
 
+// MarshalJSON implements json.Marshaller.
+//
+// json.Umarshaller is not implemented because the ID should be discarded.
```
lukechampine, line 405:
```go
// MarshalJSON implements json.Marshaller.
//
// For convenience, the transaction's ID is also calculated and included. This field is ignored during unmarshalling.
```

types/types.go
n8maninger: Nice catch. The recursion got me and I gave up to quickly.

lukechampine:
> where do we draw the line? Siacoin Siafund UTXOs and File Contracts also have IDs that are not easily exposed.

hmm, yeah. They're also more problematic to compute: you need access to the Transaction containing them. We could make `(Transaction).MarshalJSON` *really* special and override all of its children to include their ID, but that would be really ugly (as well as really slow, at least for v1 transactions -- which require you to rehash the *entire* transaction for every ID). In practice I imagine it's less common for consumers to care about those IDs anyway, though.

`Block.ID()`, though... that might be worth adding.

ChrisSchinnerl (approved):
Lgtm. We did a similar thing in `renterd` for fields that we just don't need internally since we compute them on-the-fly but having them in the API response is very nice for API consumers.

PR #233, opened 2024-11-15
Simplify RHP4 DevEx

n8maninger:
This introduces two changes to simplify client DX. These settings felt very similar to the batch size settings in RHP2 -- the host could *technically* change them, but actual renter code just assumed the default 17.83 MB was valid. This enforces a safe default assumption and reduces the complexity around hosts with differing settings.


rhp/v4/rhp.go
```diff
@@ -98,9 +107,9 @@ func (hp HostPrices) RPCReadSectorCost(length uint64) Usage {
 
 // RPCWriteSectorCost returns the cost of executing the WriteSector RPC with the
 // given sector length and duration.
-func (hp HostPrices) RPCWriteSectorCost(sectorLength uint64, duration uint64) Usage {
+func (hp HostPrices) RPCWriteSectorCost(sectorLength uint64) Usage {
```
peterjan, line 110: `nit` update docstring to reflect change

ChrisSchinnerl (dismissed):
Good stuff. That will make things less awkward.

peterjan (approved):

PR #234, opened 2024-11-18
types: Simplify SatisfiedPolicy encoding

lukechampine:
Previously, we used the policy itself to figure out how many signatures and preimages to expect. This saved two length prefixes (16 bytes), but it was annoying to implement, and also led to a vulnerability: `EncodeTo` panicked when called with the wrong number of signatures or preimages, but `UnmarshalJSON` did not validate these, so you could trigger a panic by causing a node to decode an invalid `SpendPolicy` as JSON and then encode it as binary.

The modal satisfied policy is 1 pubkey + 1 signature, so ~100 bytes, making this a ~16% size increase. Not insignificant, but tolerable.

Note that we don't need any additional validation: `(SpendPolicy).Verify` already checks for superfluous signatures and preimages.

### DIFF
--- a/types/encoding.go
+++ b/types/encoding.go
@@ -587,29 +587,10 @@ func (p SpendPolicy) EncodeTo(e *Encoder) {
 // EncodeTo implements types.EncoderTo.
 func (sp SatisfiedPolicy) EncodeTo(e *Encoder) {
 	sp.Policy.EncodeTo(e)
-	var sigi, prei int
-	var rec func(SpendPolicy)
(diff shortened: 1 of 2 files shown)


n8maninger (approved):

lukechampine:
(this will break existing consensus.dbs, btw... rip)

n8maninger:
> (this will break existing consensus.dbs, btw... rip)

Only ones that have the hard fork activated, no?

lukechampine:
oh, true. phew

PR #235, opened 2024-11-20
consensus: Rename SiafundPool -> SiafundTaxRevenue

lukechampine:
"Pool" was always a bit of a misnomer, because making a siafund claim doesn't subtract value from the "pool." Its actual function is to record the total revenue at specific points in time, so that we can calculate the delta between two points and thereby determine the value of the dividend.

I also slightly tweaked how dividends are calculated in v2. Previously, all file contract taxes were rounded down to a multiple of 10,000 H, so that they would be evenly divisible by the number of siafunds. But this isn't actually necessary; you can just divide when you calculate the claim value, and the rounding will occur then. This is slightly more fair anyway: as an extreme example, if all contract taxes had a value of 9999 H, siafund holders would never receive any dividends, no matter how many contracts were formed.

### DIFF
--- a/consensus/state.go
+++ b/consensus/state.go
@@ -370,12 +370,7 @@ func (s State) FileContractTax(fc types.FileContract) types.Currency {
 
 // V2FileContractTax computes the tax levied on a given v2 contract.
 func (s State) V2FileContractTax(fc types.V2FileContract) types.Currency {
-	sum := fc.RenterOutput.Value.Add(fc.HostOutput.Value)
-	tax := sum.Div64(25) // 4%
-	// round down to nearest multiple of SiafundCount
-	_, r := bits.Div64(0, tax.Hi, s.SiafundCount())
-	_, r = bits.Div64(r, tax.Lo, s.SiafundCount())
-	return tax.Sub(types.NewCurrency64(r))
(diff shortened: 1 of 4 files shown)


n8maninger (dismissed):

lukechampine:
I suppose we could make a separate branch for changes that should wait until the testnet is reset 🤔

ChrisSchinnerl (commented):
We should probably discuss this on a call but my initial gut feeling it to not touch the tax code because
- generally speaking I'd like to draw the line at what we have right now and not introduce any new breaking changes. There will always be stuff to tweak.
- the benefits don't outweigh the risks imo
- the fact that changing this code didn't cause any tests to fail doesn't spark joy xD

lukechampine:
Harumph. Ok, I'm fine leaving it out of this PR, but I do think we should accumulate any breaking changes in a separate branch. We'll almost certainly end up resetting Anagami once or twice before releasing the final v2 binaries; I think it'd be wasteful *not* to use those opportunities to deploy breaking changes.

btw, the reason all these consensus changes are coming now is because I'm throwing myself into the new v2 documentation, and thoroughly documenting something has a way of revealing subtle errors and inspiring new ideas. So, the sooner I finish the docs, the better. 😅

n8maninger (dismissed):

PR #237, opened 2024-11-27
Allow Foundation subsidy to be waived

lukechampine:
v2 transactions have a `NewFoundationAddress` field, which, if present, updates both the primary and failsafe addresses. This PR tweaks the update logic slightly: if the new address is `types.VoidAddress`, only the primary address is updated. `(State).FoundationSubsidy` also now returns `false` if the primary address is `types.VoidAddress`. Together, these changes allow the Foundation to effectively waive the subsidy until a new, non-void primary address is set. This is preferable to receiving and then later burning the subsidy; instead, the subsidy is simply not created in the first place.

I also renamed "primary" and "failsafe" to "subsidy" and "management" to reflect the function of these addresses in v2. The "subsidy" address receives the subsidies, while the "management" address is used to update addresses. (Note that, thanks to spend policies, the old "primary" and "failsafe" can be combined into one address using a threshold policy.)

Checking for `types.VoidAddress` might seem like a bit of a hack. Alternatively, we could expand the definition of `NewFoundationAddress`. Technically, this would change the encoding of `V2Transaction`; however, AFAICT, it would not actually cause any breakage, as the field is a pointer and has never been used in the v2 testnet (hence always encoded as `false`).

### DIFF
--- a/consensus/state.go
+++ b/consensus/state.go
@@ -257,22 +257,23 @@ func (s State) AncestorDepth() uint64 {
 }
 
 // FoundationSubsidy returns the Foundation subsidy output for the child block.
-// If no subsidy is due, the returned output has a value of zero.
 func (s State) FoundationSubsidy() (sco types.SiacoinOutput, exists bool) {
+	if s.FoundationPrimaryAddress == types.VoidAddress {
+		return types.SiacoinOutput{}, false
+	}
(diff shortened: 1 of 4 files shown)


n8maninger:
🧪

lukechampine:
Would be cool if there was a GitHub action that enforced that a test failed prior to the PR 🤔

PR #238, opened 2024-11-29
Simplify V2FileContractRenewal

lukechampine:
The tradeoff is that resolution transactions can't contain other outputs. That doesn't seem too restrictive to me, but maybe I'm forgetting a use case that requires it?

I also noticed that a renewal's `FinalRevision` wasn't being applied to the accumulator that way; instead, the accumulator would contain the most recent revision of the contract. Not a huge deal, just felt inconsistent.

### DIFF
--- a/consensus/state.go
+++ b/consensus/state.go
@@ -574,7 +574,6 @@ func (s State) ContractSigHash(fc types.V2FileContract) types.Hash256 {
 func (s State) RenewalSigHash(fcr types.V2FileContractRenewal) types.Hash256 {
 	nilSigs(
 		&fcr.NewContract.RenterSignature, &fcr.NewContract.HostSignature,
-		&fcr.FinalRevision.RenterSignature, &fcr.FinalRevision.HostSignature,
 		&fcr.RenterSignature, &fcr.HostSignature,
 	)
 	return hashAll("sig/filecontractrenewal", s.v2ReplayPrefix(), fcr)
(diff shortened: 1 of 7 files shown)


n8maninger:
This looks like it will make rhp4 require setup transactions for partial rollovers, definitely not worth it imo

n8maninger:
https://github.com/SiaFoundation/core/pull/209

lukechampine:
ah. right. drat

lukechampine:
This PR now makes the following changes:
- The new contract created by a renewal must have valid signatures. This preserves the invariant that all contracts in the accumulator have valid signatures.
- The "final revision" in a renewal has been replaced by just the final payouts (and rollover amounts). This reflects the fact that this final revision is not actually inserted in to the accumulator, nor are its signatures valid, nor does it appear as a `rev` in `ForEachV2FileContractElement` -- in other words, it's not a "real" revision at all.

I believe it's possible for the rollover values to be omitted entirely: the rollover amount can be determined by subtracting the payout values from the original contract value. However, this would require changes to RHPv4, since helpers like `RefreshCost` assume that the renter's portion of the rollover can be distinguished from the host's portion. Personally I would prefer to change RHPv4 to pass the rollover values around explicitly, so that they could be removed from `V2FileContractRenewal`; but I recognize that, like `Capacity`, it's pretty convenient to have the fields in the renewal itself.

n8maninger (approved):

PR #242, opened 2024-12-03
Fix renewal payout validation

n8maninger:
Did not notice this until after trying to update coreutils with #238.  The new renewal validation logic was checking the individual payout values against the parent file contract element. This does not account for the payouts changing in an unbroadcast revision. This changes the logic to check that the sum of the outputs is equal to the sum of the parent outputs. Also adds some additional validation tests.

### DIFF
--- a/consensus/validation.go
+++ b/consensus/validation.go
@@ -780,12 +780,24 @@ func validateV2FileContracts(ms *MidState, txn types.V2Transaction) error {
 		case *types.V2FileContractRenewal:
 			renewal := *r
 
+			if fc.RenterPublicKey != renewal.NewContract.RenterPublicKey {
+				return fmt.Errorf("file contract renewal %v changes renter public key", i)
+			} else if fc.HostPublicKey != renewal.NewContract.HostPublicKey {
+				return fmt.Errorf("file contract renewal %v changes host public key", i)
+			}
+
+			// validate that the renewal value is equal to existing contract's value.
+			// This must be done as a sum of the outputs, since the individual payouts may have
+			// changed in an unbroadcast revision.
+			totalPayout := renewal.FinalRenterOutput.Value.Add(renewal.RenterRollover).
+				Add(renewal.FinalHostOutput.Value).Add(renewal.HostRollover)
+			existingPayout := fc.RenterOutput.Value.Add(fc.HostOutput.Value)
+			if totalPayout != existingPayout {
+				return fmt.Errorf("file contract renewal %d renewal payout (%s) does not match existing contract payout %s", i, totalPayout, existingPayout)
+			}
+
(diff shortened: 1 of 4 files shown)


consensus/validation.go
```diff
@@ -780,12 +780,24 @@ func validateV2FileContracts(ms *MidState, txn types.V2Transaction) error {
 		case *types.V2FileContractRenewal:
 			renewal := *r
 
+			if fc.RenterPublicKey != renewal.NewContract.RenterPublicKey {
```
n8maninger, line 783: not strictly necessary, but the error message without it is unhelpful.

rhp/v4/rhp.go
```diff
@@ -344,6 +345,7 @@ type (
 	// RPCRenewContractSecondResponse implements Object.
 	RPCRenewContractSecondResponse struct {
 		RenterRenewalSignature  types.Signature         `json:"renterRenewalSignature"`
+		RenterContractSignature types.Signature         `json:"renterContractSignature"`
```
n8maninger, line 348: Need to pass the renter contract signature to the host now

peterjan (approved):
very much like the way the various test cases are outlined!

ChrisSchinnerl (approved):

consensus/validation.go
```diff
@@ -780,12 +780,24 @@ func validateV2FileContracts(ms *MidState, txn types.V2Transaction) error {
...
+				return fmt.Errorf("file contract renewal %v changes renter public key", i)
+			} else if fc.HostPublicKey != renewal.NewContract.HostPublicKey {
+				return fmt.Errorf("file contract renewal %v changes host public key", i)
+			}
```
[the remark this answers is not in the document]
n8maninger: We *could*, but why?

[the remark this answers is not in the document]
lukechampine: well, we allow revisions to change the keys. Granted, I don't see a ton of utility there either; key rotation is really not as much of a thing as people assume. So I'd be ok disallowing it in both cases too -- just wanna be consistent.

consensus/validation.go
```diff
@@ -780,12 +780,24 @@ func validateV2FileContracts(ms *MidState, txn types.V2Transaction) error {
...
+
+			// validate that the renewal value is equal to existing contract's value.
+			// This must be done as a sum of the outputs, since the individual payouts may have
+			// changed in an unbroadcast revision.
+			totalPayout := renewal.FinalRenterOutput.Value.Add(renewal.RenterRollover).
+				Add(renewal.FinalHostOutput.Value).Add(renewal.HostRollover)
+			existingPayout := fc.RenterOutput.Value.Add(fc.HostOutput.Value)
+			if totalPayout != existingPayout {
+				return fmt.Errorf("file contract renewal %d renewal payout (%s) does not match existing contract payout %s", i, totalPayout, existingPayout)
```
lukechampine, line 796: error is a bit inconsistent with the others, which use `(%d H)`. That might predate https://github.com/SiaFoundation/core/pull/146, so perhaps we should switch them all to `%s` now?

PR #251, opened 2024-12-10
Take storage cost into account in RenewalCost

ChrisSchinnerl:
Related to the failing tests in https://github.com/SiaFoundation/coreutils/pull/134. Broadcasting the revision fails with `siacoin inputs (600274999999999999906047180800 H) do not equal outputs (600274999999999999910241484800 H)` which seems to exactly match the storage cost. After verifying it by setting the storage price to 0, I confirmed that this change fixes the error.


n8maninger (changes requested):
Can you add some golden cases for renewal and refresh costs so we don't have to rely on the coreutils tests failing?

PR #252, opened 2024-12-10
fix(rhp4): Include storage cost in renter renewal cost

n8maninger:
Slightly different implementation of #251 that does not change the function signature

### DIFF
--- a/rhp/v4/rhp_test.go
+++ b/rhp/v4/rhp_test.go
@@ -3,6 +3,7 @@ package rhp
 import (
 	"testing"
 
+	"go.sia.tech/core/consensus"
 	"go.sia.tech/core/types"
 )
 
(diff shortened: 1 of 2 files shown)


rhp/v4/rhp_test.go
```diff
@@ -19,3 +20,212 @@ func TestMinRenterAllowance(t *testing.T) {
...
+
+			params := RPCRenewContractParams{
+				ProofHeight: renewalProofHeight,
+			}
+			tc.Modify(&contract, &params)
+
+			prices.TipHeight = renewalHeight
+			renewal, _ := RenewContract(contract, prices, params)
+			tax := cs.V2FileContractTax(renewal.NewContract)
+			renter, host := RenewalCost(cs, prices, renewal, types.ZeroCurrency)
```
ChrisSchinnerl, line 138: I'd add a hardcoded miner fee here as well. e.g. 1SC. Just to make sure it's added in the right output.

ChrisSchinnerl (changes requested):

rhp/v4/rhp_test.go
```diff
@@ -19,3 +20,212 @@ func TestMinRenterAllowance(t *testing.T) {
...
+		{
+			Description: "renewed storage - no renter rollover",
+			Modify: func(rev *types.V2FileContract, p *RPCRenewContractParams) {
+				// adjust the renewal params
+				p.Allowance = rev.RenterOutput.Value.Add(types.Siacoins(20))
+				p.Collateral = rev.TotalCollateral.Add(types.Siacoins(10))
+
+				// add storage
+				rev.Capacity = SectorSize
+				rev.Filesize = SectorSize
```
ChrisSchinnerl, line 112: Can we add another test case for capacity > filesize?

rhp/v4/rhp_test.go
```diff
@@ -19,3 +20,212 @@ func TestMinRenterAllowance(t *testing.T) {
...
+
+		renewalHeight      = 150
+		extension          = 10
+		renewalProofHeight = initialProofHeight + extension
+		renewalExpiration  = renewalProofHeight + ProofWindow
+		renewalDuration    = renewalExpiration - renewalHeight
+	)
+	cs := consensus.State{}
+	prices := HostPrices{
+		ContractPrice:   types.NewCurrency64(100),
```
ChrisSchinnerl, line 37: The tip here is implicitly set to 0. Can you set it to something > 0?

rhp/v4/rhp_test.go
```diff
@@ -19,3 +20,212 @@ func TestMinRenterAllowance(t *testing.T) {
...
+
+			prices.TipHeight = renewalHeight
+			renewal, _ := RenewContract(contract, prices, params)
+			tax := cs.V2FileContractTax(renewal.NewContract)
+			renter, host := RenewalCost(cs, prices, renewal, types.ZeroCurrency)
+			if !renter.Equals(tc.RenterCost.Add(tax)) {
+				t.Errorf("expected renter cost %v, got %v", tc.RenterCost, renter.Sub(tax))
+			} else if !host.Equals(tc.HostCost) {
+				t.Errorf("expected host cost %v, got %v", tc.HostCost, host)
+			}
```
ChrisSchinnerl, line 143: Since you already went through all the trouble to get this test to use `RenewContract`, you might as well check that the new contract's outputs match `renter + host`. Same for refresh.

ChrisSchinnerl (changes requested):

rhp/v4/rhp_test.go
```diff
@@ -19,3 +20,212 @@ func TestMinRenterAllowance(t *testing.T) {
...
+			HostCost:   types.Siacoins(10).Add(prices.Collateral.Mul64(SectorSize).Mul64(renewalDuration)),                       // collateral is calculated for the full duration
+		},
+		{
+			Description: "renewed storage - no renter rollover",
+			Modify: func(rev *types.V2FileContract, p *RPCRenewContractParams) {
+				// adjust the renewal params
+				p.Allowance = rev.RenterOutput.Value.Add(types.Siacoins(20))
+				p.Collateral = rev.TotalCollateral.Add(types.Siacoins(10))
+
+				// add storage
```
peterjan, line 110: would the test be slightly better if we start out with a non-zero file size and effectively increase it here instead of setting it?

peterjan (commented):
Looks good to me, nothing but `nits`. I'll leave it for Chris to approve

rhp/v4/rhp_test.go
n8maninger: It's changed below, so I'm not sure it matters if the initial formation height is 0.

peterjan (approved):

ChrisSchinnerl (approved):

PR #257, opened 2024-12-16
Add host key to account token

n8maninger:


lukechampine:
What's this for again?

n8maninger:
So you can’t replay account spending across hosts

lukechampine (approved):

PR #258, opened 2024-12-16
fix(consensus): Allow v1 contracts to be resolved immediately

lukechampine:
After careful testing, I found that `siad` does allow a contract to be resolved within the same block it was created. In short:
- Contracts must have a `WindowStart` of at least `currentHeight+1`
- Storage proofs check that `currentHeight` is not greater than the "trigger height," which is `WindowStart - 1`
- Therefore, if you create a contract with `WindowStart = currentHeight + 1`, its trigger height will be `currentHeight`, so a storage proof is allowed.

So in fact, you can create, revise, and resolve a v1 contract all in the same block! (On that note, the update docstrings are more descriptive now.)

Note that v2 contracts do NOT share this property. There is no such thing as an ephemeral v2 contract; that is, you can't use `types.UnassignedLeafIndex` when referencing a v2 contract. (We *could* support this, but I see little reason to.)

Lastly, I improved the validation tests by checking for specific error strings, and switched the consensus errors to use `%v` formatting for SC amounts, rather than Hastings.

### DIFF
--- a/consensus/state.go
+++ b/consensus/state.go
@@ -631,7 +631,7 @@ func (ms *MidState) fileContractElement(ts V1TransactionSupplement, id types.Fil
 }
 
 func (ms *MidState) storageProof(ts V1TransactionSupplement, id types.FileContractID) (V1StorageProofSupplement, bool) {
-	if i, ok := ms.created[id]; ok && ms.fces[i].FileContract.WindowStart == ms.base.childHeight() {
+	if i, ok := ms.created[id]; ok {
 		return V1StorageProofSupplement{FileContract: ms.fces[i], WindowID: ms.base.Index.ID}, true
 	}
 	return ts.storageProof(id)
--- a/consensus/state.go
+++ b/consensus/state.go
@@ -610,14 +610,24 @@ func (ms *MidState) siacoinElement(ts V1TransactionSupplement, id types.SiacoinO
 	if i, ok := ms.created[id]; ok {
 		return ms.sces[i], true
 	}
-	return ts.siacoinElement(id)
+	for _, sce := range ts.SiacoinInputs {
+		if sce.ID == id {
+			return sce, true
+		}
+	}
+	return types.SiacoinElement{}, false
 }
 
 func (ms *MidState) siafundElement(ts V1TransactionSupplement, id types.SiafundOutputID) (types.SiafundElement, bool) {
 	if i, ok := ms.created[id]; ok {
 		return ms.sfes[i], true
 	}
-	return ts.siafundElement(id)
+	for _, sfe := range ts.SiafundInputs {
+		if sfe.ID == id {
+			return sfe, true
+		}
+	}
+	return types.SiafundElement{}, false
 }
 
 func (ms *MidState) fileContractElement(ts V1TransactionSupplement, id types.FileContractID) (types.FileContractElement, bool) {
@@ -625,19 +635,30 @@ func (ms *MidState) fileContractElement(ts V1TransactionSupplement, id types.Fil
 		return *rev, true
 	} else if i, ok := ms.created[id]; ok {
 		return ms.fces[i], true
-	} else if rev, ok := ts.revision(id); ok {
-		return rev, true
 	}
-	sps, ok := ts.storageProof(id)
-	return sps.FileContract, ok
+	for _, fce := range ts.RevisedFileContracts {
+		if fce.ID == id {
+			return fce, true
+		}
+	}
+	for _, sps := range ts.StorageProofs {
+		if sps.FileContract.ID == id {
+			return sps.FileContract, true
+		}
+	}
+	return types.FileContractElement{}, false
 }
 
 func (ms *MidState) storageProofWindowID(ts V1TransactionSupplement, id types.FileContractID) (types.BlockID, bool) {
 	if i, ok := ms.created[id]; ok && ms.fces[i].FileContract.WindowStart == ms.base.childHeight() {
 		return ms.base.Index.ID, true
 	}
-	sps, ok := ts.storageProof(id)
-	return sps.WindowID, ok
+	for _, sps := range ts.StorageProofs {
+		if sps.FileContract.ID == id {
+			return sps.WindowID, true
+		}
+	}
+	return types.BlockID{}, false
 }
 
 func (ms *MidState) spent(id types.ElementID) (types.TransactionID, bool) {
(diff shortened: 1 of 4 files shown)


mike76-dev:
Yeah, I found it a very helpful "feature" while testing Sia-Satellite. I could form a contract and renew it straight away, without waiting.

consensus/validation.go
```diff
@@ -90,7 +90,7 @@ func ValidateOrphan(s State, b types.Block) error {
 	} else if err := validateMinerPayouts(s, b); err != nil {
 		return err
 	} else if err := validateHeader(s, b.ParentID, b.Timestamp, b.Nonce, b.ID()); err != nil {
-		return err
+		return fmt.Errorf("block has %v", err)
```
n8maninger, line 93: Should be %w

lukechampine:
> Yeah, I found it a very helpful "feature" while testing Sia-Satellite. I could form a contract and renew it straight away, without waiting.

Hmm. Perhaps we should allow "ephemeral" contracts in v2 as well, then. It would be a relatively minor change:
```diff
--- a/consensus/validation.go
+++ b/consensus/validation.go
@@ -684,6 +684,10 @@ func validateV2FileContracts(ms *MidState, txn types.V2Transaction) error {
                        return fmt.Errorf("has already been revised by contract revision %v", i)
                } else if i, ok := resolved[fce.ID]; ok {
                        return fmt.Errorf("has already been resolved by contract resolution %v", i)
+               } else if fce.StateElement.LeafIndex == types.UnassignedLeafIndex {
+                       if !ms.isCreated(fce.ID) {
+                               return errors.New("is not present in a prior transaction within this block")
+                       }
                } else if !ms.base.Elements.containsUnresolvedV2FileContractElement(fce) {
                        if ms.base.Elements.containsResolvedV2FileContractElement(fce) {
                                return errors.New("has already been resolved in a previous block")
```

 What do you think @n8maninger @ChrisSchinnerl?

n8maninger:
I prefer the v2 behavior

consensus/state.go
```diff
@@ -630,6 +630,13 @@ func (ms *MidState) fileContractElement(ts V1TransactionSupplement, id types.Fil
 	return ts.revision(id)
 }
 
+func (ms *MidState) storageProof(ts V1TransactionSupplement, id types.FileContractID) (V1StorageProofSupplement, bool) {
+	if i, ok := ms.created[id]; ok && ms.fces[i].FileContract.WindowStart == ms.base.childHeight() {
```
n8maninger, line 634: Should this return immediately if `ms.fces[i].FileContract.WindowStart == ms.base.childHeight()`?

lukechampine: not sure I follow, but we could write it this way:
```go
if ms.fces[i].FileContract.WindowStart == ms.base.childHeight() {
    if 
    return V1StorageProofSupplement{FileContract: ms.fces[i], WindowID: ms.base.Index.ID}, 
}
```
basically, there are two possibilities:
- The contract was created in an earlier block. In that case, it will be present in the transaction supplement, so we find it via `ts.storageProof`.
- The contract was created in this block. In that case, the WindowStart *must* be the next block; otherwise, we can't provide the WindowID (since it would be the ID of a future block).

lukechampine:
I suppose it's another consistency thing: if we allow siacoins and siafunds to be ephemeral, why not file contracts as well? Or: if we don't allow it for file contracts, why do we allow it for siacoins/siafunds?

consensus/state.go
n8maninger: That was basically my thought. As written, if `ok` is true and `WindowStart == ms.base.childHeight()` is false it will call `ts.storageProof` and loop through the transaction supplement which shouldn't include the contract ID since it was just created. It's unlikely to save many cycles, but the extra loop seemed unnecessary.

n8maninger:
I suppose that is true

n8maninger:
>  if we allow siacoins and siafunds to be ephemeral, why not file contracts as well?

I have a bit of an issue with this logic since contract payouts are time locked and siacoin/siafund UTXOs are not. Resolving in the same block isn't a beneficial use case. Add in complications from rollover and I'm unsure its behavior we really should support. 

Doesn't V2 use `ProofHeight` for the entropy, instead of `WindowStart-1`? If so, we wouldn't be able to have a contract that is created and resolved in the same block because ProofHeight is required to be greater than `childHeight`, right?

consensus/validation.go
```diff
@@ -351,14 +351,17 @@ func validateFileContracts(ms *MidState, txn types.Transaction, ts V1Transaction
...
+		if txid, ok := ms.spent(sp.ParentID); ok {
 			return fmt.Errorf("storage proof %v conflicts with previous proof (in %v)", i, txid)
 		}
-		sps, ok := ts.storageProof(sp.ParentID)
+		sps, ok := ms.storageProof(ts, sp.ParentID)
 		if !ok {
 			return fmt.Errorf("storage proof %v references nonexistent file contract", i)
 		}
 		fc := sps.FileContract.FileContract
+		if ms.base.childHeight() < fc.WindowStart {
```
ChrisSchinnerl, line 362: I might be missing it but this seems like it doesn't have a test covering it

consensus/validation_test.go
```diff
@@ -671,34 +712,18 @@ func TestValidateBlock(t *testing.T) {
...
-				},
-			},
-			{
-				"double-spent siafund input",
-				func(b *types.Block) {
-					txn := &b.Transactions[0]
-					txn.SiafundInputs = append(txn.SiafundInputs, txn.SiafundInputs[0])
-					txn.SiafundOutputs[0].Value += 100
-				},
-			},
```
ChrisSchinnerl, line 699: What happened to these two test cases?

ChrisSchinnerl:
I don't think there is any value in allowing file contracts to be ephemeral and "why not to be consistent" doesn't seem worthwhile to me. Contracts are meant for storing data over time and "ephemeral" means "no time" in this case. Uploading to a contract after a host knows the entropy for the storage proof is also not desirable.

consensus/validation_test.go
lukechampine: they were duplicates of earlier cases

lukechampine:
> I disagree with this logic since contract payouts are time locked and siacoin/siafund UTXOs are not.

The payouts are still timelocked; this is about the contract itself. It's the same as how siafund outputs can be ephemeral, even though a siafund input creates a timelocked claim.

> Uploading to a contract after a host knows the entropy for the storage proof is also not desirable.

That thought did cross my mind. The host could immediately throw away all segments except the selected one. On the other hand, the renter would have to be behaving really strangely in the first place to form a 0-block contract, so I feel like all bets are off.

Overall my sense is that, unlike ephemeral siacoins which have a real use case, ephemeral contracts are purely a curiosity that are *maybe* handy during testing sometimes, and useless otherwise. I'm inclined to be conservative and not change any more consensus rules.

consensus/state.go
lukechampine: I ended up always returning `true`, even if the window start is different. This is a bit odd, but it means that for early storage proofs, we return a more informative error rather than "nonexistent contract".

n8maninger: Always returning `true` doesn't seem right. If there was ever a bug that managed to call `Apply` without `Validate`, it would silently finalize the contract instead of panicking. I agree that the error is unhelpful though. Maybe `storageProof` could return an error instead?

consensus/state.go
```diff
@@ -630,6 +630,13 @@ func (ms *MidState) fileContractElement(ts V1TransactionSupplement, id types.Fil
 	return ts.revision(id)
 }
 
+func (ms *MidState) storageProof(ts V1TransactionSupplement, id types.FileContractID) (V1StorageProofSupplement, bool) {
+	if i, ok := ms.created[id]; ok {
+		return V1StorageProofSupplement{FileContract: ms.fces[i], WindowID: ms.base.Index.ID}, true
```
n8maninger, line 635: Always returning true doesn't seem right either. If there was ever a bug that managed to call Apply without Validate, it could silently finalize contracts instead of panicking. I agree that the error is unhelpful though. Maybe storageProof could return an error instead?

lukechampine: tbf there are a *lot* of things that can go wrong, loudly or silently, if you call Apply without Validate.

In this case I think the problem comes from having a single `storageProof` method instead of splitting it into `fileContract` and `windowID`. I can fix that.

n8maninger (approved):

PR #270, opened 2025-01-30
Consensus diffs

lukechampine:
This replaces the `ForEach` update API with slices of "diffs" -- new types wrapping the various element types. This was originally intended as an ergonomics improvement (since it's annoying to e.g. break out of a `ForEach` callback), but it ended up significantly simplifying most `MidState`-related code: it consolidated the interrelated maps within `MidState`, and enabled a much saner rewrite of the update JSON types.

I originally left the `ForEach` methods in place (with a `// Deprecated` warning), but later removed them entirely; we're going to update all the callsites in `coreutils` anyway, so there's little reason to keep them around. (`ForEachTreeNode` remains, though, since it's used by `explored`.)

### DIFF
--- a/consensus/application.go
+++ b/consensus/application.go
@@ -735,46 +735,6 @@ func (au ApplyUpdate) UpdateElementProof(e *types.StateElement) {
 	au.eau.updateElementProof(e)
 }
 
-// ForEachSiacoinElement calls fn on each siacoin element related to the applied
-// block.
-//
-// Deprecated: Use SiacoinElements instead.
-func (au ApplyUpdate) ForEachSiacoinElement(fn func(sce types.SiacoinElement, created, spent bool)) {
-	for _, sce := range au.sces {
-		fn(sce.SiacoinElement, sce.Created, sce.Spent)
-	}
-}
-
-// ForEachSiafundElement calls fn on each siafund element related to the applied
-// block.
-//
-// Deprecated: Use SiafundElements instead.
-func (au ApplyUpdate) ForEachSiafundElement(fn func(sfe types.SiafundElement, created, spent bool)) {
-	for _, sfe := range au.sfes {
-		fn(sfe.SiafundElement, sfe.Created, sfe.Spent)
-	}
-}
-
-// ForEachFileContractElement calls fn on each file contract element related to
-// the applied block.
-//
-// Deprecated: Use FileContractElements instead.
-func (au ApplyUpdate) ForEachFileContractElement(fn func(fce types.FileContractElement, created bool, rev *types.FileContractElement, resolved, valid bool)) {
-	for _, fce := range au.fces {
-		fn(fce.FileContractElement, fce.Created, fce.Revision, fce.Resolved, fce.Valid)
-	}
-}
-
-// ForEachV2FileContractElement calls fn on each v2 file contract element
-// related to the applied block.
-//
-// Deprecated: Use V2FileContractElements instead.
-func (au ApplyUpdate) ForEachV2FileContractElement(fn func(fce types.V2FileContractElement, created bool, rev *types.V2FileContractElement, res types.V2FileContractResolutionType)) {
-	for _, fce := range au.v2fces {
-		fn(fce.V2FileContractElement, fce.Created, fce.Revision, fce.Resolution)
-	}
-}
-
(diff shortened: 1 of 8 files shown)


consensus/application.go
```diff
@@ -608,119 +644,88 @@ func (ms *MidState) ApplyBlock(b types.Block, bs V1BlockSupplement) {
...
+	v2fces []V2FileContractElementDiff
+	aes    []types.AttestationElement
+	cie    types.ChainIndexElement
+
 	eau elementApplyUpdate
 }
 
+// SiacoinElementDiffs returns the siacoin element diffs related to the applied
+// block.
+func (au ApplyUpdate) SiacoinElementDiffs() []SiacoinElementDiff { return au.sces }
```
n8maninger, line 704: Is there a benefit to wrapping the field access in a method rather than exposing the slice directly?

consensus/application.go
```diff
@@ -339,124 +341,156 @@ func dupProof(se *types.StateElement) {
 	se.MerkleProof = append([]types.Hash256(nil), se.MerkleProof...)
 }
 
-func (ms *MidState) addSiacoinElement(id types.SiacoinOutputID, sco types.SiacoinOutput) {
-	sce := types.SiacoinElement{
+func (ms *MidState) recordSiacoinElement(id types.SiacoinOutputID) *SiacoinElementDiff {
```
peterjan, line 344: `nit` `ensureSiacoinElement` would've made it slightly easier to understand

peterjan: Doesn't it make sense to inline it and get rid of `recordX` all together? In `ApplyTransaction` for instance, we fetch the element, panic if it doesn't exist, then spends it which 'll record it first. I noticed `ApplyV2Transaction` does this slightly differently by the way, as it doesn't check for existence and doesn't panic.

Is it possible to end up with a `SiacoinElementDiff` where `spent` is true but `created` is false? I guess my real question is: "In what scenario does spending an element effectively create a new element when it calls record". The record almost feels like it hides an inconsistency, or an order of events type thing where we would spent and create within the same block but not necessarily in the correct order or something.

This might be desired behaviour but figured I'd point it out to be sure.

consensus/application.go
```diff
@@ -339,124 +341,156 @@ func dupProof(se *types.StateElement) {
...
+		fced.Revision = &rev
 	}
 }
 
 func (ms *MidState) resolveV2FileContractElement(fce types.V2FileContractElement, res types.V2FileContractResolutionType, txid types.TransactionID) {
-	ms.v2res[fce.ID] = res
-	ms.spends[fce.ID] = txid
+	fced := ms.recordV2FileContractElement(fce.ID)
+	if fced.Created {
+		panic("consensus: resolved a newly-created v2 contract")
```
peterjan, line 483: Do we ever update the `created` bool of a fce in the midstate? I'm sure it's fine but since it's a newly introduced check (and panic) I wanted to draw attention to it to be 100% sure it's valid.

peterjan (commented):
left some questions but LGTM 👍

consensus/application.go
lukechampine: I considered it, but methods let you define an `Update` interface, and give us some flexibility if we end up changing the internal representation again

n8maninger (approved):

consensus/application.go
lukechampine: I'm not too averse to inlining it; mostly this is about eliding the "does this element already exist" check in a few places. With regard to siacoin and siafund elements, it doesn't add a ton of value (it saves one check, in `spendSiacoinElement`). It's more helpful for file contract elements, because it saves two checks (in `revise` and `resolve`).

> Is it possible to end up with a SiacoinElementDiff where spent is true but created is false? I guess my real question is: "In what scenario does spending an element effectively create a new element when it calls record".

If an element was created in a previous block, and spent in this block, then we want the first block to produce a diff with `Created: true, Spent: false` and the second block to produce `Created: false, Spent: true`. Even though the second block didn't create the element, it still *affected* it, so we need to record that.

consensus/application.go
lukechampine: no; `Created` is a field of t`FileContractElementDiff`, not `FileContractElement`, and once it's set to `true` we never touch it again.

PR #274, opened 2025-02-11
Add helpers to get revision as an element from v1/v2 file contract element diffs

chris124567:
See https://github.com/SiaFoundation/explored/pull/169#discussion_r1950507575

### DIFF
--- a/.changeset/add_helpers_to_get_revision_as_an_element_from_v1v2_file_contract_element_diffs.md
+++ b/.changeset/add_helpers_to_get_revision_as_an_element_from_v1v2_file_contract_element_diffs.md
@@ -1,0 +1,9 @@
+---
+default: minor
+---
+
+# Add helpers to get revision as an element from v1/v2 file contract element diffs
+
+#274 by @chris124567
+
+See https://github.com/SiaFoundation/explored/pull/169#discussion_r1950507575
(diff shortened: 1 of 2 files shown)


consensus/state.go
```diff
@@ -615,6 +615,18 @@ type FileContractElementDiff struct {
 	Valid    bool                `json:"valid"`
 }
 
+// RevisionElement returns the revision, if present, as a FileContractElement.
+func (diff FileContractElementDiff) RevisionElement() *types.FileContractElement {
+	if diff.Revision == nil {
```
n8maninger, line 620: Would rather return `(FileContractElement, bool)` or `(FileContractElement, error)` here over a pointer.

.changeset/add_helpers_to_get_revision_as_an_element_from_v1v2_file_contract_element_diffs.md
```diff
@@ -0,0 +1,9 @@
+---
+default: minor
+---
+
+# Add helpers to get revision as an element from v1/v2 file contract element diffs
+
+#274 by @chris124567
+
+See https://github.com/SiaFoundation/explored/pull/169#discussion_r1950507575
```
n8maninger, line 9: Distill the comment here instead of linking to it

n8maninger (approved):

PR #276, opened 2025-02-14
Add RHP4 RPC replenish types

n8maninger:
Adds an RPC to RHP4 that enables renters to set a target balance for all accounts instead of first fetching the current balances then funding the accounts with the difference. This is primarily to speed up account funding and reduce round trips when managing a large number of accounts.

### DIFF
--- a/rhp/v4/rhp.go
+++ b/rhp/v4/rhp.go
@@ -474,7 +474,7 @@ type (
 	}
 	// RPCReplenishAccountsResponse is the response type for RPCReplenishAccounts.
 	RPCReplenishAccountsResponse struct {
-		Cost types.Currency `json:"cost"`
+		Deposits []AccountDeposit `json:"deposits"`
 	}
 	// RPCReplenishAccountsSecondResponse is the second response type for RPCReplenishAccounts.
 	RPCReplenishAccountsSecondResponse struct {
@@ -588,6 +588,14 @@ func (r *RPCReplenishAccountsRequest) ChallengeSigHash(revisionNumber uint64) ty
 // ValidChallengeSignature checks the challenge signature for validity.
 func (r *RPCReplenishAccountsRequest) ValidChallengeSignature(fc types.V2FileContract) bool {
 	return fc.RenterPublicKey.VerifyHash(r.ChallengeSigHash(fc.RevisionNumber), r.ChallengeSignature)
+}
+
+// TotalCost returns the total cost to the renter of the replenish RPC.
+func (r *RPCReplenishAccountsResponse) TotalCost() (total types.Currency) {
+	for _, deposit := range r.Deposits {
+		total = total.Add(deposit.Amount)
+	}
+	return
 }
 
 // NewContract creates a new file contract with the given settings.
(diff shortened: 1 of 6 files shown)


rhp/v4/rhp.go
```diff
@@ -458,6 +465,26 @@ type (
...
+	// RPCReplenishAccountsRequest is the request type for RPCReplenishAccounts.
+	RPCReplenishAccountsRequest struct {
+		Accounts           []Account            `json:"accounts"`
+		Target             types.Currency       `json:"target"`
+		ContractID         types.FileContractID `json:"contractID"`
+		ChallengeSignature types.Signature      `json:"challengeSignature"`
+	}
+	// RPCReplenishAccountsResponse is the response type for RPCReplenishAccounts.
+	RPCReplenishAccountsResponse struct {
+		Cost types.Currency `json:"cost"`
```
ChrisSchinnerl, line 477: Thinking about it a bit more, do we want to return more info here? This is definitely the most efficient way to design the protocol, but there might be information we need that we can't get from it right now. e.g. how much we actually fund a specific account with. On the other hand it's also a features to not have detailed insights into the spending habits of users.

rhp/v4/rhp.go
```diff
@@ -547,6 +574,22 @@ func (r *RPCRefreshContractRequest) ValidChallengeSignature(existing types.V2Fil
 	return existing.RenterPublicKey.VerifyHash(r.ChallengeSigHash(existing.RevisionNumber), r.ChallengeSignature)
 }
 
+// ChallengeSigHash returns the hash of the challenge signature used for
+// signing.
+func (r *RPCReplenishAccountsRequest) ChallengeSigHash(revisionNumber uint64) types.Hash256 {
```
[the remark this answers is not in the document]
n8maninger: Pretty much. Theoretically you could lock the contract and stir up a bunch of IO before the host notices. This lets the host detect and unlock the contract immediately.

rhp/v4/rhp.go
n8maninger: If we decide to return the specific account, we can also change it to fund as many accounts as possible instead of erroring.

rhp/v4/rhp.go
```diff
@@ -458,6 +465,26 @@ type (
...
+		Deposits []AccountDeposit `json:"deposits"`
+	}
+	// RPCReplenishAccountsSecondResponse is the second response type for RPCReplenishAccounts.
+	RPCReplenishAccountsSecondResponse struct {
+		RenterSignature types.Signature `json:"renterSignature"`
+	}
+	// RPCReplenishAccountsThirdResponse is the third response type for RPCReplenishAccounts.
+	RPCReplenishAccountsThirdResponse struct {
+		HostSignature types.Signature `json:"hostSignature"`
+	}
```
lukechampine, line 486: could the host send their signature along with the deposits?

n8maninger: It would be more efficient, but I believe we want to be consistent with the host sending their signature last since they are primarily responsible for maintaining and broadcasting the state.

lukechampine (approved):

peterjan (approved):

ChrisSchinnerl (approved):

PR #284, opened 2025-03-07
Fix `maxLen`

peterjan:
There are several instances where `maxLen` returns a size that doesn't take into account a transaction being part of the request. While updating I noticed some values diverged from their impl probably due to changes being made to the request/response structs over time. I took them along with me but the important ones are the `reasonableTransactionSetSize` ones of course.

I found out because my v2 migration on my node, which was chugging along nicely, halted all of a sudden. Contracts were no longer being refreshed, resulting in empty accounts, resulting in failed migrations. One thing that's puzzling me a bit though is the error message: "couldn't renew/refresh contract: host responded with error: failed to read host inputs response: failed to read request: EOF ". That error message seems impossible, I double checked and I still can't explain how you could ever get failed to read response, failed to read request.

### DIFF
--- a/rhp/v4/encoding.go
+++ b/rhp/v4/encoding.go
@@ -180,7 +180,7 @@ func (r *RPCFormContractRequest) decodeFrom(d *types.Decoder) {
 	types.DecodeSlice(d, &r.RenterParents)
 }
 func (r *RPCFormContractRequest) maxLen() int {
-	return reasonableObjectSize
+	return reasonableTransactionSetSize
 }
 
 func (r *RPCFormContractResponse) encodeTo(e *types.Encoder) {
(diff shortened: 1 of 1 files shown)


rhp/v4/encoding.go
```diff
@@ -663,7 +663,7 @@ func (r *RPCFundAccountsResponse) decodeFrom(d *types.Decoder) {
 	r.HostSignature.DecodeFrom(d)
 }
 func (r *RPCFundAccountsResponse) maxLen() int {
-	return 8 + (sizeofCurrency * MaxAccountBatchSize) + sizeofSignature
+	return (sizeofCurrency * MaxAccountBatchSize) + sizeofSignature
```
peterjan, line 666: ignore these, reverting

ChrisSchinnerl (approved):

ChrisSchinnerl:
> One thing that's puzzling me a bit though is the error message: "couldn't renew/refresh contract: host responded with error: failed to read host inputs response: failed to read request: EOF ". That error message seems impossible, I double checked and I still can't explain how you could ever get failed to read response, failed to read request.

That's because the host is getting the EOF when reading the request and returns it to the renter. The renter reads the response, gets the error instead of the inputs and then reports that the host responded with an error when we tried to read the inputs response and the error the host sent is EOF.

lukechampine (approved):

PR #290, opened 2025-03-16
consensus: fix contract revision aliasing

n8maninger:


consensus/state.go
```diff
@@ -700,10 +700,9 @@ func (ms *MidState) siafundElement(ts V1TransactionSupplement, id types.SiafundO
 
 func (ms *MidState) fileContractElement(ts V1TransactionSupplement, id types.FileContractID) (types.FileContractElement, bool) {
 	if i, ok := ms.elements[id]; ok {
-		if ms.fces[i].Revision != nil {
-			fce := ms.fces[i].FileContractElement
-			fce.FileContract = *ms.fces[i].Revision
-			return fce, true
+		rev, ok := ms.fces[i].RevisionElement()
```
n8maninger, line 703: This isn't strictly necessary, but we have the helper regardless

consensus/application.go
```diff
@@ -432,6 +432,7 @@ func (ms *MidState) reviseFileContractElement(fce types.FileContractElement, rev
 	} else if fced.Revision != nil {
 		*fced.Revision = rev
 	} else {
+		dupProof(&fce.StateElement)
```
lukechampine, line 435: never could I possibly have predicted that Merkle proof aliasing would bite me (again (again))

n8maninger: Should definitely just go ahead and `dupProof` anywhere with a Merkle tree regardless of if it's currently an issue lol.

ChrisSchinnerl (approved):
Lgtm. I assume you verified this by syncing manually. Maybe @lukechampine can follow this up with a regression test now that we know what happened.

peterjan (dismissed):
In `spendSiafundElement` and `spendSiacoinElement` we call `dupProof` after assigning the element. I'd move that up so it's consistent with all other spots.

ChrisSchinnerl (approved):

PR #292, opened 2025-03-18
all: Add Element memory management methods and linter

lukechampine:
This should go a long way towards catching memory aliasing bugs like https://github.com/SiaFoundation/core/pull/290. I should have done this ages ago. 😬 Basically, the linter flags anything like:
```go
sce := txn.SiacoinInputs[0].Parent // "shallow copy of SiacoinElement; use Move, Share, or Copy"
cau.UpdateElementProof(&sce)
```
To silence it, you annotate each copy with one of those ownership methods, and then we panic if you try to call `UpdateElementProof` on an element that you don't exclusively own.

Annotating every copy is kind of annoying, but it wasn't too onerous to churn through them once the linter was working. In case it's not clear, the idea is:
- Default to `Copy`.
- If you're not going to retain the element (e.g. if you're a function that just inspects elements and returns them), you can use `Share`. Putting the element inside of a struct (like a transaction) counts as retaining it! Usually, it's okay to use `Share` when passing an element to a function (and consequently, if you're a function that takes an element, you should treat it as shared).
- If you "own" the element and you just need to tweak it for whatever reason, you can use `Move`. This is most common when calling `UpdateElementProof` on each element in a map (since can't get a pointer to map values, you need to make a temporary copy, update that, then store it back in the map). `Move` is like `Share`, except you "retain ownership" so you're still allowed to update the proof.

One nice thing about this approach is that it's opt-in. If you're importing `core`, you don't *need* to be aware of any of this stuff, even if you're directly touching elements. But you are *encouraged* to run the linter on your code and annotate all your copies. I've done this with `coreutils` and it was pretty easy.

The biggest shortcoming is that this won't detect shallow copies of structs containing an element. For example, if you copy a transaction without `.DeepCopy`, it won't complain. Flagging every such copy would be impractical, since you would need to add `Copy`/`Share`/`Move` methods to *every* type containing an element.

Lastly -- yes, this is (to paraphrase Greenspun) "an ad hoc, informally-specified, bug-ridden, slow implementation of half of Rust."

### DIFF
--- a/consensus/application.go
+++ b/consensus/application.go
@@ -337,10 +337,6 @@ func ApplyOrphan(s State, b types.Block, targetTimestamp time.Time) State {
 	return next
 }
 
-func dupProof(se *types.StateElement) {
-	se.MerkleProof = append([]types.Hash256(nil), se.MerkleProof...)
-}
-
 func (ms *MidState) recordSiacoinElement(id types.SiacoinOutputID) *SiacoinElementDiff {
 	if i, ok := ms.elements[id]; ok {
 		return &ms.sces[i]
@@ -358,7 +354,6 @@ func (ms *MidState) createSiacoinElement(id types.SiacoinOutputID, sco types.Sia
 		SiacoinOutput: sco,
 	}
 	sced.Created = true
-	dupProof(&sced.SiacoinElement.StateElement)
 	return sced
 }
 
@@ -370,9 +365,8 @@ func (ms *MidState) createImmatureSiacoinElement(id types.SiacoinOutputID, sco t
 
 func (ms *MidState) spendSiacoinElement(sce types.SiacoinElement, txid types.TransactionID) {
 	sced := ms.recordSiacoinElement(sce.ID)
-	sced.SiacoinElement = sce
+	sced.SiacoinElement = sce.Copy()
 	sced.Spent = true
-	dupProof(&sced.SiacoinElement.StateElement)
 	ms.spends[sce.ID] = txid
 }
 
@@ -399,9 +393,8 @@ func (ms *MidState) createSiafundElement(id types.SiafundOutputID, sfo types.Sia
 
 func (ms *MidState) spendSiafundElement(sfe types.SiafundElement, txid types.TransactionID) {
 	sfed := ms.recordSiafundElement(sfe.ID)
-	sfed.SiafundElement = sfe
+	sfed.SiafundElement = sfe.Copy()
 	sfed.Spent = true
-	dupProof(&sfed.SiafundElement.StateElement)
 	ms.spends[sfe.ID] = txid
 }
 
@@ -422,7 +415,6 @@ func (ms *MidState) createFileContractElement(id types.FileContractID, fc types.
 		FileContract: fc,
 	}
 	fced.Created = true
-	dupProof(&fced.FileContractElement.StateElement)
 	ms.siafundTaxRevenue = ms.siafundTaxRevenue.Add(ms.base.FileContractTax(fc))
 }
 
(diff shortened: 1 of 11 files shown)


types/types.go
```diff
@@ -1239,6 +1231,175 @@ func (res *V2FileContractResolution) UnmarshalJSON(b []byte) error {
...
+		jsonTxn
+	}{txn.ID(), jsonTxn(txn)})
+}
+
+// To guard against memory ownership bugs, all Element types have Move, Share,
+// and Copy methods. This enables a linter to flag any instances where Element
+// memory is not explicitly managed.
+
+// CanMutate returns whether the element's memory can safely be mutated.
+func (se StateElement) CanMutate() bool { return !se.shared }
```
ChrisSchinnerl, line 1255: All of this looks promising but this makes me a bit sad. It means you still need to think about calling `CanMutate` before mutating which is something you might forget.

The only way around this is probably to have a container type around StateElement.

```
type StateElementContainer struct {
  stateElement *StateElement
  shared bool
}

func (c *StateElementContainer) MutRef() *StateElement {
  if c.shared {
    panic("bad stuff")
  }
  return c.stateElement
}
```

In this case you wouldn't even be able to get a `StateElement` to mutate without calling `MutRef` to get a mutable reference. Still not foolproof and probably 10 times more nasty.

lukechampine: `CanMutate` only really exists to be used by `UpdateElementProof`. I suppose it might not be needed, though; the `Move` method is kinda like your `MutRef`, in that it will panic if called on a shared element. So we can do:
```go
func UpdateElementProof(e *types.StateElement) {
    _ = e.Move() // panics if e is shared
    // ...
}
```
The panic message won't be as nice, but it's still an improvement over having a separate method just for one caller.

consensus/application.go
```diff
@@ -957,7 +942,7 @@ func (au *ApplyUpdate) UnmarshalJSON(b []byte) error {
 	au.fces = js.FileContractElementDiffs
 	au.v2fces = js.V2FileContractElementDiffs
 	au.aes = js.AttestationElements
-	au.cie = js.ChainIndexElement
+	au.cie = js.ChainIndexElement.Move()
```
peterjan, line 945: Don't we have to call `Move` in a for loop on anything with a state element here?

peterjan (approved):
Looks good to me! Tested a sync from scratch and it doesn't panic 😅 . Something I don't fully understand is that if we produced a bad block at one point, wouldn't going back through all walletd versions, swapping out core for this, highlight the issue and panic when it adds the block that caused it?

lukechampine:
possibly, but you would need to update the `walletd` code as well (to call `Move`/`Share`/`Copy` everywhere) -- feasible if you only have to do it once, but not so much if you need to do it for every version you want to test.

ChrisSchinnerl:
So this looks promising but when I tried to sync a node on Zen it got stuck at height 114.771.
Siascan is at 115,183.

image

I'll try to resync again but this is a bit odd considering that we have successfully synced multiple nodes earlier this week.

ChrisSchinnerl:
Sync number 3 is stuck again it seems. Same block.

image

lukechampine:
I reached 114,771 as well. After that, the node will occasionally hear about a longer chain and try to reorg to it, but fails on a commitment hash mismatch at block `113529::5ef176ed`. So I think the longest Zen chain still has an invalid element tree. At this point, the only fix is for everyone (particularly miners) to update and then either resync or run `MigrateDB`.

lukechampine:
btw, this PR shouldn't have had any effect on the contents of the element tree. I suppose we should double-check that, though.

PR #294, opened 2025-03-19
gateway: Fix missing SendCheckpoint RPC

lukechampine:
🙃


n8maninger:
🧪

lukechampine:
Does this count? https://github.com/SiaFoundation/coreutils/pull/202

n8maninger:
Nope

lukechampine:
😭

What sort of test did you have in mind? This bug would have been caught by a test that checked every input for `ObjectForID`... but that test would *also* need to enumerate all of the IDs, so it would be just as vulnerable to the same bug.

n8maninger (approved):

PR #299, opened 2025-04-17
Add MaxHostCollateral

ChrisSchinnerl:
This PR adds `MaxHostCollateral` as a counterpart to `MinRenterAllowance`.

It also removes the `duration` since it doesn't affect the outcome of the functions since the ratio stays the same regardless of the duration.

### DIFF
--- a/.changeset/add_maxhostcollateral_helper.md
+++ b/.changeset/add_maxhostcollateral_helper.md
@@ -1,0 +1,5 @@
+---
+default: patch
+---
+
+# Add MaxHostCollateral helper
--- a/rhp/v4/rhp.go
+++ b/rhp/v4/rhp.go
@@ -773,6 +773,9 @@ func ReviseForReplenish(fc types.V2FileContract, amount types.Currency) (types.V
 // MinRenterAllowance returns the minimum allowance required to justify the given
 // host collateral.
 func MinRenterAllowance(hp HostPrices, collateral types.Currency) types.Currency {
+	if hp.Collateral.IsZero() {
+		return types.ZeroCurrency
+	}
 	maxCollateralBytes := collateral.Div(hp.Collateral)
 	return hp.StoragePrice.Mul(maxCollateralBytes)
 }
@@ -780,6 +783,9 @@ func MinRenterAllowance(hp HostPrices, collateral types.Currency) types.Currency
 // MaxHostCollateral returns the maximum amount of collateral a host can justify
 // to put into a contract for the given allowance.
 func MaxHostCollateral(hp HostPrices, allowance types.Currency) types.Currency {
+	if hp.StoragePrice.IsZero() {
+		return types.MaxCurrency
+	}
 	maxCollateralBytes := allowance.Div(hp.StoragePrice)
 	return hp.Collateral.Mul(maxCollateralBytes)
 }
--- a/rhp/v4/rhp.go
+++ b/rhp/v4/rhp.go
@@ -772,9 +772,16 @@ func ReviseForReplenish(fc types.V2FileContract, amount types.Currency) (types.V
 
 // MinRenterAllowance returns the minimum allowance required to justify the given
 // host collateral.
-func MinRenterAllowance(hp HostPrices, duration uint64, collateral types.Currency) types.Currency {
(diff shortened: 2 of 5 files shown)


rhp/v4/validation.go
```diff
@@ -130,7 +130,7 @@ func (req *RPCFormContractRequest) Validate(pk types.PublicKey, tip types.ChainI
 	duration := expirationHeight - hp.TipHeight
 	// calculate the minimum allowance required for the contract based on the
 	// host's locked collateral and the contract duration
-	minRenterAllowance := MinRenterAllowance(hp, duration, req.Contract.Collateral)
+	minRenterAllowance := MinRenterAllowance(hp, req.Contract.Collateral)
```
peterjan, line 133: update comment x3

rhp/v4/rhp.go
```diff
@@ -772,9 +772,16 @@ func ReviseForReplenish(fc types.V2FileContract, amount types.Currency) (types.V
 
 // MinRenterAllowance returns the minimum allowance required to justify the given
 // host collateral.
-func MinRenterAllowance(hp HostPrices, duration uint64, collateral types.Currency) types.Currency {
-	maxCollateralBytes := collateral.Div(hp.Collateral).Div64(duration)
-	return hp.StoragePrice.Mul64(duration).Mul(maxCollateralBytes)
+func MinRenterAllowance(hp HostPrices, collateral types.Currency) types.Currency {
+	maxCollateralBytes := collateral.Div(hp.Collateral)
```
peterjan, line 776: should we add a sanity check for the division by zero

rhp/v4/rhp.go
```diff
@@ -772,9 +772,16 @@ func ReviseForReplenish(fc types.V2FileContract, amount types.Currency) (types.V
...
+	return hp.StoragePrice.Mul(maxCollateralBytes)
+}
+
+// MaxHostCollateral returns the maximum amount of collateral a host can justify
+// to put into a contract for the given allowance.
+func MaxHostCollateral(hp HostPrices, allowance types.Currency) types.Currency {
+	maxCollateralBytes := allowance.Div(hp.StoragePrice)
```
peterjan, line 783: same here

rhp/v4/validation.go
```diff
@@ -170,7 +170,7 @@ func (req *RPCRenewContractRequest) Validate(pk types.PublicKey, tip types.Chain
 	duration := expirationHeight - hp.TipHeight
```
peterjan, line 170: potential underflow here

rhp/v4/validation.go
```diff
@@ -170,7 +170,7 @@ func (req *RPCRenewContractRequest) Validate(pk types.PublicKey, tip types.Chain
...
 	// calculate the minimum allowance required for the contract based on the
 	// host's locked collateral and the contract duration
-	minRenterAllowance := MinRenterAllowance(hp, duration, req.Renewal.Collateral)
+	minRenterAllowance := MinRenterAllowance(hp, req.Renewal.Collateral)
```
peterjan, line 173: `nit` we can move up this check `case req.Renewal.Allowance.IsZero():` and it add to the request validation

rhp/v4/validation.go
```diff
@@ -167,6 +170,9 @@ func (req *RPCRenewContractRequest) Validate(pk types.PublicKey, tip types.Chain
 	// validate the contract fields
 	hp := req.Prices
 	expirationHeight := req.Renewal.ProofHeight + ProofWindow
+	if expirationHeight < hp.TipHeight {
```
peterjan, line 173: Probably the answer is no, but should we/could we check the `req.Renewal.ProofHeight` against the `TipHeight` instead? Would be stricter, just not sure if we have to allow renewing/refreshing within the proof window.

ChrisSchinnerl: Actually I noticed we already have that check further down so I reverted this.

.changeset/add_maxhostcollateral_helper.md
```diff
@@ -0,0 +1,5 @@
+---
+default: patch
```
n8maninger, line 2: Add another changeset with a breaking change for removing the duration parameter.

PR #301, opened 2025-04-24
Clean up validation

n8maninger:
Reverts #298 to the old behavior and removes unused params.

Looking at the original PR, #298, I think the logic change there is actually incorrect. It's not clear to me why a contract refresh would ever have plenty of allowance but not enough collateral since the funds always flow from the renter to the host. 

The subtest that got changed in https://github.com/SiaFoundation/coreutils/pull/215 should have been a valid state based on the intended behavior.

The desired behavior is that the host's total collateral locked is a consistent ratio with the renter's total spending. In the case of that subtest, the renter has locked `110 SC` total and the host has locked `220 SC` total. With the new logic, that is now considered a bad state because the renter spent `25 SC` of its allowance funding an account, giving it `85 SC` remaining allowance to `220 SC` collateral. The renter ends up needing to lock significantly more SC to compensate for the account spending.

### DIFF
--- a/rhp/v4/rhp.go
+++ b/rhp/v4/rhp.go
@@ -16,6 +16,10 @@ const (
 	// the contract expires.
 	ProofWindow = 144 // 24 hours
 
+	// MinContractDuration is the minimum duration, in blocks, a host will
+	// accept for a contract.
+	MinContractDuration = 18 // 3 hours
+
 	// TempSectorDuration is the number of blocks that temp sectors are expected to be stored
 	// before being removed
 	TempSectorDuration = 144 * 3
@@ -680,14 +684,15 @@ func NewContract(p HostPrices, cp RPCFormContractParams, hostKey types.PublicKey
 }
 
 // ContractCost calculates the cost to the host and renter for forming a contract.
-func ContractCost(cs consensus.State, p HostPrices, fc types.V2FileContract, minerFee types.Currency) (renter, host types.Currency) {
-	renter = fc.RenterOutput.Value.Add(p.ContractPrice).Add(minerFee).Add(cs.V2FileContractTax(fc))
+func ContractCost(cs consensus.State, fc types.V2FileContract, minerFee types.Currency) (renter, host types.Currency) {
+	contractCost := fc.HostOutput.Value.Sub(fc.TotalCollateral) // (contract price + locked collateral) - locked collateral
+	renter = fc.RenterOutput.Value.Add(contractCost).Add(minerFee).Add(cs.V2FileContractTax(fc))
 	host = fc.TotalCollateral
 	return
 }
 
 // RenewalCost calculates the cost to the host and renter for renewing a contract.
-func RenewalCost(cs consensus.State, p HostPrices, r types.V2FileContractRenewal, minerFee types.Currency) (renter, host types.Currency) {
+func RenewalCost(cs consensus.State, r types.V2FileContractRenewal, minerFee types.Currency) (renter, host types.Currency) {
 	contractCost := r.NewContract.HostOutput.Value.Sub(r.NewContract.TotalCollateral) // (contract price + storage cost + locked collateral) - locked collateral
 	renter = r.NewContract.RenterOutput.Value.Add(contractCost).Add(minerFee).Add(cs.V2FileContractTax(r.NewContract)).Sub(r.RenterRollover)
 	host = r.NewContract.TotalCollateral.Sub(r.HostRollover)
@@ -696,10 +701,11 @@ func RenewalCost(cs consensus.State, p HostPrices, r types.V2FileContractRenewal
 
 // RefreshCost calculates the cost to the host and renter for refreshing a contract.
 func RefreshCost(cs consensus.State, p HostPrices, r types.V2FileContractRenewal, minerFee types.Currency) (renter, host types.Currency) {
+	// Unlike renewals, it is not possible to calculate the renter's cost of the contract during refreshes without the host's prices
+	// because the existing renter and host outputs are rolled into the new contract to ensure the existing data
+	// is still covered by collateral.
 	renter = r.NewContract.RenterOutput.Value.Add(p.ContractPrice).Add(minerFee).Add(cs.V2FileContractTax(r.NewContract)).Sub(r.RenterRollover)
-	// the calculation is different from renewal because the host's revenue is also rolled into the refresh.
-	// This calculates the new collateral the host is expected to put up:
-	// new collateral = (new revenue + existing revenue + new collateral + existing collateral) - new revenue - (existing revenue + existing collateral)
+	// new collateral = (existing revenue + contract price + existing revenue + new collateral + existing collateral) - contract price - (existing revenue + existing collateral)
 	host = r.NewContract.HostOutput.Value.Sub(p.ContractPrice).Sub(r.HostRollover)
 	return
 }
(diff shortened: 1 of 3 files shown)


rhp/v4/validation.go
```diff
@@ -149,84 +153,83 @@ func (req *RPCFormContractRequest) Validate(pk types.PublicKey, tip types.ChainI
...
 	case req.MinerFee.IsZero():
 		return errors.New("miner fee must be greater than 0")
 	case req.Basis == (types.ChainIndex{}):
 		return errors.New("basis must be set")
-	case req.Renewal.ProofHeight <= existingProofHeight:
-		return fmt.Errorf("renewal proof height must be greater than existing proof height %v", existingProofHeight)
+	case req.Renewal.ProofHeight <= existing.ProofHeight:
+		return fmt.Errorf("renewal proof height must be greater than existing proof height %v", existing.ProofHeight)
+	case req.Renewal.ProofHeight < minProofHeight:
+		return fmt.Errorf("renewal proof height must be greater than %v", minProofHeight)
```
peterjan, line 174: `nit` must be at least, considering we allow if it's equal

rhp/v4/validation.go
```diff
@@ -149,84 +153,83 @@ func (req *RPCFormContractRequest) Validate(pk types.PublicKey, tip types.ChainI
...
 	}
 
 	// validate the request fields
+	minProofHeight := minProofHeight(tip, req.Prices)
 	switch {
 	case req.MinerFee.IsZero():
 		return errors.New("miner fee must be greater than 0")
 	case req.Basis == (types.ChainIndex{}):
 		return errors.New("basis must be set")
+	case existing.ProofHeight <= minProofHeight:
```
peterjan, line 215: shouldn't this be strictly less, considering the docs on `minProofHeight` indicate it's the minimum height necessary for the host to accept the contract

rhp/v4/validation.go
```diff
@@ -149,84 +153,83 @@ func (req *RPCFormContractRequest) Validate(pk types.PublicKey, tip types.ChainI
...
 	// validate the contract fields
 	hp := req.Prices
 	// calculate the minimum allowance required for the contract based on the
 	// host's locked collateral and the contract duration
-	postRefreshAllowance := req.Refresh.Allowance.Add(existingAllowance)
-	postRefreshCollateral := req.Refresh.Collateral.Add(existingCollateral)
-	minRenterAllowance := MinRenterAllowance(hp, postRefreshCollateral)
-	// refreshes add collateral on top of the existing collateral
-	totalCollateral := req.Refresh.Collateral.Add(existingTotalCollateral)
+	minRenterAllowance := MinRenterAllowance(hp, req.Refresh.Collateral)
```
peterjan, line 223: `nit` update the comment, the duration no longer influences the min allowance

rhp/v4/rhp.go
```diff
@@ -696,10 +701,11 @@ func RenewalCost(cs consensus.State, p HostPrices, r types.V2FileContractRenewal
...
 // RefreshCost calculates the cost to the host and renter for refreshing a contract.
 func RefreshCost(cs consensus.State, p HostPrices, r types.V2FileContractRenewal, minerFee types.Currency) (renter, host types.Currency) {
+	// Unlike renewals, it is not possible to calculate the renter's cost of the contract during refreshes without the host's prices
+	// because the existing renter and host outputs are rolled into the new contract to ensure the existing data
+	// is still covered by collateral.
 	renter = r.NewContract.RenterOutput.Value.Add(p.ContractPrice).Add(minerFee).Add(cs.V2FileContractTax(r.NewContract)).Sub(r.RenterRollover)
-	// the calculation is different from renewal because the host's revenue is also rolled into the refresh.
-	// This calculates the new collateral the host is expected to put up:
-	// new collateral = (new revenue + existing revenue + new collateral + existing collateral) - new revenue - (existing revenue + existing collateral)
+	// new collateral = (existing revenue + contract price + existing revenue + new collateral + existing collateral) - contract price - (existing revenue + existing collateral)
```
peterjan, line 708: `r.NewContract.HostOutput.Value` is the new and existing revenue and collateral plus the contract price right? we have existing revenue twice here 😅

rhp/v4/rhp.go
```diff
@@ -16,6 +16,10 @@ const (
 	// the contract expires.
 	ProofWindow = 144 // 24 hours
 
+	// MinContractDuration is the minimum duration, in blocks, a host will
+	// accept for a contract.
+	MinContractDuration = 18 // 3 hours
```
peterjan, line 21: seems very low but I guess it doesn't matter

rhp/v4/rhp.go
```diff
@@ -680,14 +684,15 @@ func NewContract(p HostPrices, cp RPCFormContractParams, hostKey types.PublicKey
 }
 
 // ContractCost calculates the cost to the host and renter for forming a contract.
-func ContractCost(cs consensus.State, p HostPrices, fc types.V2FileContract, minerFee types.Currency) (renter, host types.Currency) {
-	renter = fc.RenterOutput.Value.Add(p.ContractPrice).Add(minerFee).Add(cs.V2FileContractTax(fc))
+func ContractCost(cs consensus.State, fc types.V2FileContract, minerFee types.Currency) (renter, host types.Currency) {
+	contractCost := fc.HostOutput.Value.Sub(fc.TotalCollateral) // (contract price + locked collateral) - locked collateral
```
ChrisSchinnerl, line 688: This makes me nervous. I feel like it should be `.Sub(fc.MissedHostValue)` but I guess when forming a contract `MissedHostValue` and `TotalCollateral` are equal.

rhp/v4/validation.go
```diff
@@ -114,6 +114,9 @@ func (req *RPCFormContractRequest) Validate(pk types.PublicKey, tip types.ChainI
 		return fmt.Errorf("prices are invalid: %w", err)
 	}
 
+	minProofHeight := minProofHeight(tip, req.Prices)
+	maxExpirationHeight := max(tip.Height, req.Prices.TipHeight) + maxDuration
```
ChrisSchinnerl, line 118: use your helper

rhp/v4/validation.go
```diff
@@ -300,3 +303,17 @@ func (req *RPCReplenishAccountsRequest) Validate() error {
...
+// has enough time to be confirmed before the proof window begins.
+func minProofHeight(tip types.ChainIndex, hp HostPrices) uint64 {
+	return max(tip.Height, hp.TipHeight) + MinContractDuration
+}
+
+// maxExpirationHeight returns the maximum expiration height for a
+// contract. This is the maximum expiration height which a host will
+// accept in a contract.
+func maxExpirationHeight(tip types.ChainIndex, hp HostPrices, maxDuration uint64) uint64 {
+	return max(tip.Height, hp.TipHeight) + maxDuration
```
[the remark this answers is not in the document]
n8maninger: The thought was either the renter or the host was not synced. Honestly, we should probably just use `TipHeight` as the base for duration like we do for every other RPC.

peterjan (approved):

rhp/v4/validation.go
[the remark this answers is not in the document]
ChrisSchinnerl: I agree. Hosts should be careful to not hand out price tables which are valid forever to avoid issues but then also use the TipHeight. Otherwise we are just watering down the purpose of the field and reintroduce issues we had in RHP2.

PR #305, opened 2025-05-13
Add address field

n8maninger:
Similar to the `id` field on Transactions, adds `address` to the json representation of Siacoin and Siafund inputs for ease of use in other languages.


n8maninger:
Am I gonna get away with this? 😅

ChrisSchinnerl (approved):
Love it. We are doing something like that in `renterd` too for the UI. 
Only nit I have is that we could have unit tests to make sure the field is present. e.g. by comparing against a hardcoded string.

PR #307, opened 2025-05-14
Add UTXO id to siacoin and siafund json marshalling

n8maninger:
Adds the UTXO ID to the Siacoin and Siafund output JSON for convenience.


n8maninger:
@alexfreska the JSON encoding for Siacoin and Siafund outputs and inputs will have additional convenience fields. The types in the SDK should be updated to reflect this change. We'll be pushing it to all our nodes shortly.


Please verify, but it should look like this:

Input
```json
{
  // regular fields
  address: "asdasdad"
}
```

Output
```json
{
    // regular fields
    id: "asdasdasd"
}
```

PR #311, opened 2025-05-20
consensus: Change Commitment to a Merkle root

lukechampine:
This restores compatibility with the Stratum mining protocol. Stratum servers simply need to include an additional leaf hash when computing the Merkle tree:
```go
hashAll(uint8(0), "commitment", s.v2ReplayPrefix(), types.Hash256(hashAll(s)), minerAddr)
```

Of course, once the RequireHeight is reached and v1 transactions are no longer allowed, they will also need to adjust the `Coinbase1` and `Coinbase2` values they send, to account for the different encoding of v2 transactions.

### DIFF
--- a/consensus/state.go
+++ b/consensus/state.go
@@ -530,10 +530,11 @@ func (s State) PartialSigHash(txn types.Transaction, cf types.CoveredFields) typ
 	return h.Sum()
 }
 
-// TransactionsCommitment returns the commitment hash covering the transactions
-// that comprise a child block.
-func (s State) TransactionsCommitment(txns []types.Transaction, v2txns []types.V2Transaction) types.Hash256 {
+// Commitment computes the commitment hash for a child block with the given
+// transactions and miner address.
+func (s State) Commitment(minerAddr types.Address, txns []types.Transaction, v2txns []types.V2Transaction) types.Hash256 {
 	var acc blake2b.Accumulator
+	acc.AddLeaf(hashAll(uint8(0), "commitment", s.v2ReplayPrefix(), types.Hash256(hashAll(s)), minerAddr))
 	for _, txn := range txns {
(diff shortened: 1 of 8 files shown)


consensus/state.go
```diff
@@ -530,10 +530,11 @@ func (s State) PartialSigHash(txn types.Transaction, cf types.CoveredFields) typ
...
 
-// TransactionsCommitment returns the commitment hash covering the transactions
-// that comprise a child block.
-func (s State) TransactionsCommitment(txns []types.Transaction, v2txns []types.V2Transaction) types.Hash256 {
+// Commitment computes the commitment hash for a child block with the given
+// transactions and miner address.
+func (s State) Commitment(minerAddr types.Address, txns []types.Transaction, v2txns []types.V2Transaction) types.Hash256 {
 	var acc blake2b.Accumulator
+	acc.AddLeaf(hashAll(uint8(0), "commitment", s.v2ReplayPrefix(), types.Hash256(hashAll(s)), minerAddr))
 	for _, txn := range txns {
```
n8maninger, line 538: Do we need to commit to the miner payouts?

n8maninger:
I added regression tests for v1 and v2 to ensure the stratum logic we're aware of continues to work.

consensus/state.go
lukechampine: good question! v2 blocks only have a single payout, and its value is fully determined by the block height, so there is no need to commit to it.

n8maninger: I did a double take because it’s still a slice. Sneaky https://github.com/SiaFoundation/core/blob/498585cae108f0532f333bbd85e5b74ee0df7471/consensus/validation.go#L58

ChrisSchinnerl (approved):

PR #312, opened 2025-05-21
internal/blake2b -> blake2b

n8maninger:


lukechampine:
but why 🤔

n8maninger:
Cause I keep wanting the accumulator and I’m tired of copying it 😅

PR #315, opened 2025-05-26
Fix missing leaf prefix

n8maninger:


ChrisSchinnerl (approved):

lukechampine:
Added helper methods. I'm tempted to put unexported copies of them into `consensus` and `gateway` instead, but I'm guessing mining pool code will want them...?

n8maninger:
> Added helper methods. I'm tempted to put unexported copies of them into `consensus` and `gateway` instead, but I'm guessing mining pool code will want them...?

Yeah, we'll want them exported

PR #322, opened 2025-06-16
Remove support for v1 peers

ChrisSchinnerl:
Since the v2 hardfork has activated and everyone who doesn't run a recent version of `core` can't participate in the network anymore, I think we can remove support for v1 as additional hardening.


lukechampine (approved):
farewell `smux`, rest in peace alongside `muxado` 🫡

PR #328, opened 2025-06-26
Fix resolution json

n8mgr:
```
cannot unmarshal object into Go struct field ApplyUpdate.applied.update.v2FileContractElementDiffs.resolution of type types.V2FileContractResolutionType.
```

Fixes the JSON decoding without breaking compatibility


peterjan (approved):
LGTM 🔥

PR #332, opened 2025-07-11
v2: The Final Cut

lukechampine:
The difficulty adjustment algorithm in v2 was intended to be a straight port of v1, merely swapping "target" for "difficulty" and cleaning up some historic cruft. Unfortunately, it also introduced a bug, which has led to increased variance in block times. This caused the v2 require height to be reached earlier than expected. Block times have slowly normalized, but until the bug is fixed, there is no guarantee they will remain normal.

After reviewing the v1 algorithm, I have opted to change the squared error term to a proportional error, with the same target of "10,000 seconds of drift = 10 seconds of shift." Squaring the error means that, when approaching zero, the adjustment shrinks faster than the error does, meaning we never quite reach zero error. It also does not correct large errors more quickly, as the 0.4% limit precludes any large adjustments anyway.

I have run a thorough suite of simulations on the three adjustment algorithms, and this one is the clear winner. It smoothly corrects the existing drift, without triggering large spikes or dips in difficulty, and it fully attenuates to zero instead of asymptotically approaching it.

Changing the algorithm requires a hardfork, so I decided to include two other small fixes alongside this one. First, v2 blocks no longer include the value of their miner payout. v2 blocks are already restricted to a single payout, and there is only one legal value for that payout: the block reward plus the sum of the transaction fees. It is good practice to remove multiple sources of truth; there is one less piece of information to verify. Second, the commitment hash no longer includes deprecated State fields, namely `Depth`, `ChildTarget`, and `OakTarget` (which are equivalent to `TotalWork`, `Difficulty`, and `OakWork`, respectively). Covering these fields meant that all nodes had to accurately maintain them, for no reason. These changes mainly benefit v2-native nodes; it allows them to use different representations of the `Block` and `State` types.

We can decide on the mainnet `FinalCutHeight` in the subsequent coreutils PR. I think 60 days is reasonable, but due to high block time variance, I'll need to run more simulations to accurately map that duration to a block height.

### DIFF
--- a/consensus/state.go
+++ b/consensus/state.go
@@ -123,13 +123,13 @@ type State struct {
 
 	Index             types.ChainIndex `json:"index"`
 	PrevTimestamps    [11]time.Time    `json:"prevTimestamps"` // newest -> oldest
-	Depth             types.BlockID    `json:"depth"`
-	ChildTarget       types.BlockID    `json:"childTarget"`
+	Depth             types.BlockID    `json:"depth"`          // DEPRECATED: use TotalWork
+	ChildTarget       types.BlockID    `json:"childTarget"`    // DEPRECATED: use PoWTarget or Difficulty
 	SiafundTaxRevenue types.Currency   `json:"siafundTaxRevenue"`
 
 	// Oak hardfork state
 	OakTime   time.Duration `json:"oakTime"`
-	OakTarget types.BlockID `json:"oakTarget"`
+	OakTarget types.BlockID `json:"oakTarget"` // DEPRECATED: use OakWork
 	// Foundation hardfork state
 	FoundationSubsidyAddress    types.Address `json:"foundationSubsidyAddress"`
 	FoundationManagementAddress types.Address `json:"foundationManagementAddress"`
@@ -247,6 +247,14 @@ func (s State) BlockReward() types.Currency {
 		return s.Network.MinimumCoinbase
 	}
 	return r
+}
+
+// PoWTarget returns the proof-of-work target for the child block.
+func (s State) PoWTarget() types.BlockID {
+	if s.childHeight() < s.Network.HardforkV2.FinalCutHeight {
+		return s.ChildTarget
+	}
+	return invTarget(s.Difficulty.n)
 }
 
 // MaturityHeight is the height at which various outputs created in the child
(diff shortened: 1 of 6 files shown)


consensus/state.go
```diff
@@ -248,6 +249,14 @@ func (s State) BlockReward() types.Currency {
 	return r
 }
 
+// PoWTarget returns the proof-of-work target for the child block.
+func (s State) PoWTarget() types.BlockID {
+	if s.childHeight() < s.Network.HardforkV2.FinalCutHeight {
+		return s.ChildTarget
+	}
+	return invTarget(s.Difficulty.n)
```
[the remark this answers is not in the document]
lukechampine: Prior to v2, `s.ChildTarget` and `invTarget(s.Difficulty.n)` can be (very slightly) different due to rounding.

consensus/validation.go
```diff
@@ -63,19 +63,22 @@ func validateMinerPayouts(s State, b types.Block) error {
...
-		if overflow {
-			return errors.New("miner payouts overflow")
+		if sum != expectedSum {
+			return fmt.Errorf("miner payout sum (%v) does not match block reward + fees (%v)", sum, expectedSum)
 		}
-	}
-	if sum != expectedSum {
-		return fmt.Errorf("miner payout sum (%v) does not match block reward + fees (%v)", sum, expectedSum)
+	} else if !b.MinerPayouts[0].Value.IsZero() {
+		return errors.New("miner payout must be zero after finalcut hardfork")
```
lukechampine, line 81: if we want to avoid making any extra work for miners/pools, this change probably isn't tenable :/

lukechampine: decided to relax it to "the payout value _may_ be omitted." Hopefully in the long run we'll observe that everyone is omitting the value, and then we can have an uncontroversial fork that enforces the omission.

for context, the main reason I want to get rid of this (besides it being redundant) is that, even in v2 blocks, the miner payout values use the v1 currency encoding. In fact, they are the _only_ place in v2 that uses the old encoding, so it annoys me that we have to maintain this one special case. At this point, we can't change the block encoding itself, but we _can_ require the value to be omitted entirely, which accomplishes the same goal of enabling us to completely strip out the v1 currency code.

n8maninger:
Missing a major changeset

PR #336, opened 2025-07-17
Add version to RPCs

n8mgr:
Changes the last byte of the RPC ID to a version byte. Preserves backwards compatibility since existing RPCs already have 0.

The reason for the change is a desire to change the renter rollover logic of RPCRefresh. Currently, we add the allowance on top of the remaining allowance and roll everything into the new contract. This is different from renewals and formations where the `Allowance` param is the value the renter wants for their allowance.

We could add a new specifier instead, but this felt better than the potential of having a bunch of `types.NewSpecifier("RPCRefresh2").

### DIFF
--- a/rhp/v4/rhp.go
+++ b/rhp/v4/rhp.go
@@ -38,55 +38,50 @@ const (
 )
 
 // An RPCSpecifier uniquely identifies an RPC.
-type RPCSpecifier [15]byte
-
-// An RPCHeader pairs an RPC's specifier with a version number.
-type RPCHeader struct {
-	ID      RPCSpecifier
-	Version uint8
-}
+type RPCSpecifier types.Specifier
 
 // String implements fmt.Stringer.
-func (rh RPCHeader) String() string {
-	b := string(bytes.TrimRight(rh.ID[:], "\x00"))
+func (rs RPCSpecifier) String() string {
+	b := string(bytes.TrimRight(rs[:15], "\x00"))
 	for _, c := range b {
 		if !(('A' <= c && c <= 'Z') || ('a' <= c && c <= 'z') || ('0' <= c && c <= '9')) {
 			return strconv.Quote(b)
 		}
 	}
-	return fmt.Sprintf("%s.%d", b, rh.Version)
+	return fmt.Sprintf("%s.%d", b, rs[15])
+}
+
+func rpcSpecifier(name string, version uint8) (s RPCSpecifier) {
+	// last byte is the version, ensure the name is at most 15 bytes
+	if len(name) > len(s)-1 {
+		panic(fmt.Sprintf("ID too long: len(%q) > 15", name))
+	}
+	copy(s[:15], name)
+	s[15] = version
+	return
 }
 
 // RPC identifiers.
 var (
-	RPCAccountBalanceID    = rpcSpecifier("AccountBalance")
-	RPCFundAccountsID      = rpcSpecifier("FundAccounts")
-	RPCReplenishAccountsID = rpcSpecifier("ReplAccounts")
+	RPCAccountBalanceID    = rpcSpecifier("AccountBalance", 0)
+	RPCFundAccountsID      = rpcSpecifier("FundAccounts", 0)
+	RPCReplenishAccountsID = rpcSpecifier("ReplAccounts", 0)
 
-	RPCAppendSectorsID = rpcSpecifier("AppendSectors")
-	RPCFreeSectorsID   = rpcSpecifier("FreeSectors")
-	RPCSectorRootsID   = rpcSpecifier("SectorRoots")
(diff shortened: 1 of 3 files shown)


n8maninger:
@lukechampine this PR is effectively doing what you're suggesting, but instead of appending to the string it uses the last byte. Would this have been "more acceptable?"

```go
// An RPCSpecifier uniquely identifies an RPC.
type RPCSpecifier types.Specifier

// String implements fmt.Stringer.
func (rs RPCSpecifier) String() string {
	b := string(bytes.TrimRight(rs[:15], "\x00"))
	for _, c := range b {
		if !(('A' <= c && c <= 'Z') || ('a' <= c && c <= 'z') || ('0' <= c && c <= '9')) {
			return strconv.Quote(b)
		}
	}
	return fmt.Sprintf("%s.%d", b, rs[15])
}

func rpcSpecifier(name string, version uint8) (s RPCSpecifier) {
	// last byte is the version, ensure the name is at most 15 bytes
	if len(name) > len(s)-1 {
		panic(fmt.Sprintf("ID too long: len(%q) > 15", name))
	}
	copy(s[:15], name)
	s[15] = version
	return
}
```

I don't really feel too strongly about appending to the string instead. Most of https://github.com/SiaFoundation/coreutils/pull/299 are ergonomic changes to automatically select which RPC to use. If we feel strongly about explicitly splitting the RPCs and having the user decide which to call, we can get rid of most of the diff in that PR.

n8maninger:
I went ahead and pushed that up. If you'd still rather go with appending to the string, let me know.

n8maninger:
And here's the coreutils version with a minimal diff https://github.com/SiaFoundation/coreutils/pull/300

rhp/v4/rhp.go
```diff
@@ -36,25 +37,50 @@ const (
...
 
+// An RPCSpecifier uniquely identifies an RPC.
+type RPCSpecifier types.Specifier
+
+// String implements fmt.Stringer.
+func (rs RPCSpecifier) String() string {
+	b := string(bytes.TrimRight(rs[:15], "\x00"))
+	for _, c := range b {
+		if !(('A' <= c && c <= 'Z') || ('a' <= c && c <= 'z') || ('0' <= c && c <= '9')) {
+			return strconv.Quote(b)
```
ChrisSchinnerl, line 48: Should we just not allow anything exotic?

rhp/v4/rhp.go
```diff
@@ -36,25 +37,50 @@ const (
...
+
+func rpcSpecifier(name string, version uint8) (s RPCSpecifier) {
+	// last byte is the version, ensure the name is at most 15 bytes
+	if len(name) > len(s)-1 {
+		panic(fmt.Sprintf("ID too long: len(%q) > 15", name))
+	}
+	copy(s[:15], name)
+	s[15] = version
+	return
+}
```
ChrisSchinnerl, line 62: Short test to compare a created specifier to its raw bytes would be nice

rhp/v4/transport.go
```diff
@@ -18,16 +18,16 @@ func withDecoder(r io.Reader, maxLen int, fn func(*types.Decoder)) error {
 	return d.Err()
 }
 
-// ReadID reads an RPC ID from the stream.
-func ReadID(r io.Reader) (id types.Specifier, err error) {
-	err = withDecoder(r, 16, id.DecodeFrom)
+// ReadID reads an RPC header from the stream.
+func ReadID(r io.Reader) (id RPCSpecifier, err error) {
+	err = withDecoder(r, 16, (*types.Specifier)(&id).DecodeFrom)
 	return
```
ChrisSchinnerl, line 24: Shouldn't we also check that the combination of RPC and version is a valid one here? 
It's usually a good idea to only construct and return valid instances of a type when parsing data.

peterjan (approved):

rhp/v4/transport.go
ChrisSchinnerl: Tbh it doesn't matter too much. The whole `transport.go` file should probably live in `coreutils` anyway since it is a utility file that is neither used nor tested in `core`. But I'd argue that if we don't perform validation, it would be better to return `types.Specifier` here. If I call `ParseHTTPMethod` of some `http-core` library I'd also expect an error if it is a non-std method. 

Makes me wish this was Rust. Because then `RPCSpecifier` would probably be an enum wrapping a `Specifier` and to return it we'd be forced to validate it here. Which would then nicely force all devs to exhaustively switch over all possible variations whenever they update the dependency.

ChrisSchinnerl (approved):

rhp/v4/transport.go
n8maninger: You love to watch the world burn don't you?

PR #337, opened 2025-07-17
Split RPC refresh logic

n8mgr:
Previously, renters had to add to their remaining allowance when refreshing a contract. Now, renters can set the allowance of the new contract explicitly without forcing an increase. This brings RPCRefresh in line with the behavior of RPCForm and RPCRenew.

Requires #336

### DIFF
--- a/.changeset/add_rpcrefreshpartial_id.md
+++ b/.changeset/add_rpcrefreshpartial_id.md
@@ -1,8 +1,0 @@
----
-default: minor
----
-
-# Added RPCRefreshPartial ID
-
-This adds the ID for RPCRefreshPartial a new RPC to make contract refreshes 
-more efficient for both renters and hosts.
--- a/.changeset/added_new_version_of_rpcrefresh.md
+++ b/.changeset/added_new_version_of_rpcrefresh.md
@@ -1,9 +1,7 @@
 ---
-default: major
+default: minor
 ---
 
-# Changed RPCRefresh to allow renters to explicitly set the desired spending for refreshed contracts.
+# Added RPCRefreshPartial ID and RefreshContractPartialRollover helper
 
-Previously, renters had to add to their remaining allowance when refreshing a contract. Now, 
-renters can set the allowance of the new contract explicitly without forcing an increase. This
-brings RPCRefresh in line with the behavior of RPCForm and RPCRenew.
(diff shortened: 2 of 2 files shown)


rhp/v4/rhp.go
```diff
@@ -892,8 +892,49 @@ func RenewContract(fc types.V2FileContract, prices HostPrices, rp RPCRenewContra
 	}
 }
 
-// RefreshContract creates a contract renewal for the refresh RPC.
-func RefreshContract(fc types.V2FileContract, prices HostPrices, rp RPCRefreshContractParams) (types.V2FileContractRenewal, Usage) {
+// RefreshContractPartialRollover creates a contract renewal for the refresh RPC v1.
+func RefreshContractPartialRollover(fc types.V2FileContract, prices HostPrices, rp RPCRefreshContractParams) (types.V2FileContractRenewal, Usage) {
+	var renewal types.V2FileContractRenewal
+	renewal.FinalRenterOutput = fc.RenterOutput
```
peterjan, line 898: sSetting the final renter and host output here is redundant because they are both overwritten. If we do this for a reason though, for clarity perhaps, does it make sense to rewrite:

```go
if fc.RenterOutput.Value.Cmp(rp.Allowance) > 0 {
    renewal.RenterRollover = rp.Allowance
} else {
    renewal.RenterRollover = fc.RenterOutput.Value
}
renewal.FinalRenterOutput.Value = renewal.FinalRenterOutput.Value.Sub(renewal.RenterRollover)
```

in a way that we use `renewal.FinalRenterOutput` instead of `fc.RenterOutput`. I don't think it does but then we should probably get rid of these two assignments at the top of the function.

rhp/v4/rhp.go
ChrisSchinnerl, line 1: I feel like we are missing a change to `validation.go` unless we expect `Validate` to behave the same for both partial and full refreshes?

Right now it compares whether the ratio between added allowance and added collateral is sane for a full refresh.
For a partial refresh it checks whether the ratio between the set allowance and added collateral is sane.

So if you refresh in a loop, I think you can just multiple the collateral in a contract right now without adding more allowance.

rhp/v4/rhp.go
n8maninger: The renter and host output `Value` field is overwritten. The `Address` field is not. We could set both separately, but I think that makes it more likely we introduce a bug.

rhp/v4/rhp.go
n8maninger: This should be irrelevant if the math checks out now.

n8maninger:
Rebased this off master so we can merge cleanly since #336 seems significantly more controversial

ChrisSchinnerl (approved):

peterjan (approved):

PR #340, opened 2025-07-25
`Account.Token` -> `NewToken`

chris124567:
This method does not use any of the fields on `Account`

### DIFF
--- a/rhp/v4/rhp.go
+++ b/rhp/v4/rhp.go
@@ -262,9 +262,9 @@ func (a *Account) UnmarshalText(b []byte) error {
 	return nil
 }
 
-// Token returns a signed account token authorizing spending from the account on the
-// host.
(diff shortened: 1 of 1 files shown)


.changeset/accounttoken_newaccounttoken.md
```diff
@@ -0,0 +1,9 @@
+---
+default: major
+---
+
+# `Account.Token` -> `NewAccountToken`
+
+#340 by @chris124567
+
+This method does not use any of the fields on `Account`
```
ChrisSchinnerl, line 9: Auto generated changeset. Better to just do "Change `Account.Token` method to `NewAccountToken` function" or similar.

ChrisSchinnerl (changes requested):

ChrisSchinnerl (approved):

PR #342, opened 2025-07-28
Document `V2FileContract` type and add useful helpers for revenue calculations

ChrisSchinnerl:
Since the revenue related contract fields are not the easiest to wrap you head around I added some documentation on the `V2FileContract` type as well as helpers for some frequently done math. 

I also added `RemainingAllowance` and `RemainingCollateral` which are more or less just aliases but imo they still make contracts nicer to use. The fields `MissedHostValue` and `RenterOutput` are technically correct but make more sense when looked at in the context of the consensus code and the methods `RemainingCollateral` and `RemainingAllowance` feel more intuitive when used in the context of the RHP.

### DIFF
--- a/rhp/v4/rhp.go
+++ b/rhp/v4/rhp.go
@@ -858,7 +858,7 @@ func RenewContract(fc types.V2FileContract, prices HostPrices, rp RPCRenewContra
 	return renewal, Usage{
 		RPC:              prices.ContractPrice,
 		Storage:          renewal.NewContract.HostOutput.Value.Sub(renewal.NewContract.TotalCollateral).Sub(prices.ContractPrice),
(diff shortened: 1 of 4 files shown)


rhp/v4/rhp.go
```diff
@@ -877,8 +877,8 @@ func RefreshContractPartialRollover(fc types.V2FileContract, prices HostPrices,
 	// the host output needs to cover the existing risked collateral,
 	// existing revenue, and the new collateral to ensure the existing data
 	// is still protected.
-	hostRevenueRisked := fc.HostOutput.Value.Sub(fc.MissedHostValue)
-	hostRiskedCollateral := fc.TotalCollateral.Sub(fc.MissedHostValue)
+	hostRevenueRisked := fc.RiskedHostRevenue()
```
n8maninger, line 880: `hostRevenue + hostRiskedCollateral`

PR #395, opened 2026-04-09
increase size of RPCFormContractResponse

ChrisSchinnerl:
We have been inconsistent with the `maxLen` of request and response types including either `RenterInputs` or `HostInputs`. Most of them (well the requests) already used `reasonableTransactionSetSize` but some still used the smaller reasonableObjectSize`.

I found out because the indexer kept failing to renew a contract with a host due to `UnexpectedEOF` failures caused by the limit reader being limited to a size too small to contain the `RPCRenewContractResponse`.

After deploying a branch with an increased size the contract renewed successfully. 

NOTE: Alternatively we could also stop being fancy, remove `maxLen` and just use a default limit of 5MiB or so.

### DIFF
--- a/.changeset/increase_maxlen_for_request_and_responset_types_including_renterinputs_or_hostinputs.md
+++ b/.changeset/increase_maxlen_for_request_and_responset_types_including_renterinputs_or_hostinputs.md
@@ -2,4 +2,4 @@
 default: patch
 ---
 
-# Increase maxLen for request and responset types including RenterInputs or HostInputs.
+# Increase maxLen for request and response types including RenterInputs or HostInputs.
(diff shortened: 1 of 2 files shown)


.changeset/increase_maxlen_for_request_and_responset_types_including_renterinputs_or_hostinputs.md
```diff
@@ -0,0 +1,5 @@
+---
+default: patch
+---
+
+# Increase maxLen for request and responset types including RenterInputs or HostInputs.
```
n8mgr: sp, but looks good

n8mgr (dismissed):

lukechampine (dismissed):

PR #397, opened 2026-04-21
Pass in host address when creating contract for refresh/renewal

ChrisSchinnerl:


rhp/v4/rhp_test.go
```diff
@@ -403,21 +411,23 @@ func TestRefreshPartialRolloverCost(t *testing.T) {
...
+			var hostAddressNew types.Address
+			frand.Read(hostAddressNew[:])
 			contract, _ := NewContract(prices, RPCFormContractParams{
 				RenterPublicKey: renterKey,
 				RenterAddress:   types.StandardAddress(renterKey),
 				Allowance:       types.Siacoins(300),
 				Collateral:      types.Siacoins(400),
 				ProofHeight:     initialProofHeight,
-			}, hostKey, types.StandardAddress(hostKey))
+			}, hostKey, hostAddressNew)
```
peterjan: Why do you pass the new address here? The test is only meaningful if you pass the new address when you do the rollover right

peterjan (changes requested):

rhp/v4/rhp_test.go
ChrisSchinnerl: Probably because I started typing `hostAddress` and hit autocomplete without looking whether it says `New` or `Old` at the end. When I reverted it back to `StandardAddress` from `hostAddressOld` the string search didn't find it.

peterjan (approved):

n8mgr (approved):

PR #401, opened 2026-04-28
Add RHP4 pool RPCs and types for shared balance pools

n8mgr:
Pools let a renter deposit once into a shared backing pool and have multiple accounts draw from it on demand, reducing the per-account allowance needed and the total capital tied up in idle balances. This is especially useful for multi-tenant setups where one party holds the funds and many account keypairs spend from them.

### DIFF
--- a/rhp/v4/rhp.go
+++ b/rhp/v4/rhp.go
@@ -653,6 +653,7 @@ type (
 	RPCAttachPoolsRequest struct {
 		Attachments []PoolAttachment `json:"attachments"`
 	}
+	// RPCAttachPoolsResponse an empty success response
 	RPCAttachPoolsResponse struct{}
 
 	// RPCDetachPoolsRequest batches one or more detachments. Applied
@@ -660,6 +661,7 @@ type (
 	RPCDetachPoolsRequest struct {
 		Detachments []PoolDetachment `json:"detachments"`
 	}
+	// RPCDetachPoolsResponse an empty success response
 	RPCDetachPoolsResponse struct{}
 )
 
--- a/rhp/v4/rhp.go
+++ b/rhp/v4/rhp.go
@@ -58,6 +58,11 @@ var (
 	RPCWriteSectorID  = types.NewSpecifier("WriteSector")
 	RPCVerifySectorID = types.NewSpecifier("VerifySector")
 	RPCSettingsID     = types.NewSpecifier("Settings")
+
+	RPCFundPoolsID      = types.NewSpecifier("FundPools")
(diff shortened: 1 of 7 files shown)


rhp/v4/rhp.go
```diff
@@ -622,6 +627,42 @@ type (
 		Balances      []types.Currency `json:"balances"`
 		HostSignature types.Signature  `json:"hostSignature"`
 	}
+
+	// PoolAttachment authorizes attaching a single account to a pool. The
+	// signature is by the pool's private key, since the pool holds funds at
+	// risk. Each entry is independently signed so a batch can mix pools owned
```
chris124567, line 632: Won't an indexer just own one pool with a bunch of accounts attached/why would people with different pools not just make separate requests?

n8mgr: We’re planning on one pool per user so each one being a potentially separate pool is beneficial and harmless (albeit verbose) if we switch later.

rhp/v4/rhp.go
```diff
@@ -58,6 +58,11 @@ var (
 	RPCWriteSectorID  = types.NewSpecifier("WriteSector")
 	RPCVerifySectorID = types.NewSpecifier("VerifySector")
 	RPCSettingsID     = types.NewSpecifier("Settings")
+
+	RPCFundPoolsID      = types.NewSpecifier("FundPools")
```
peterjan: Are we intentionally reusing the fund/replenish account request types? Should we make that explicit in core by doing something like `RPCFundPoolsRequest = RPCFundAccountsRequest`? It's minor but I only realised what was happening when looking at coreutils, could be useful to document that here.

rhp/v4/rhp.go
```diff
@@ -704,6 +745,48 @@ func (r *RPCReplenishAccountsResponse) TotalCost() (total types.Currency) {
...
+	hostKey.EncodeTo(h.E)
+	a.Account.EncodeTo(h.E)
+	a.Pool.EncodeTo(h.E)
+	h.E.WriteTime(a.ValidUntil)
+	return h.Sum()
+}
+
+// ValidSignature checks that the signature was produced by the pool's
+// private key against this host's public key.
+func (a *PoolAttachment) ValidSignature(hostKey types.PublicKey) bool {
```
peterjan, line 763: I think it's worth adding signature test coverage for `PoolAttachment` and `PoolDetachment`

peterjan (dismissed):
Left two suggestions but otherwise LGTM 👍

ChrisSchinnerl (approved):

PR #427, opened 2026-09-15
Merkle tree algorithm optimizations

lukechampine:
These improvements were found by Astra (Ultra).

> `addLeaves`: Every growing original tree shares a proof suffix with the first added leaf. Recording its first sibling during insertion lets us copy that suffix afterward, eliminating the per-merge scan of all 64 trees.
> `updateLeaves`: Recurse only where updated leaf indices diverge. Both branches contain updates; proofRoot handles intervening paths. This removes empty-branch cases and subtree-boundary arithmetic, including its overflow at large indices.
>
> Local benchmarks show carry-heavy insertions 33–48% faster and single updates about 95% faster. The tradeoff is roughly 3% slower large, dense updates.


ChrisSchinnerl (commented):
Could you add some benchmarks to verify the improvements?
Also do we expect this to have significant impact on the renter host protocol or sync speeds?

lukechampine:
Benchmark figures:

Optimization | Speedup
-- | --
addLeaves | 1.02–1.49×
updateLeaves, 256 updates | 1.01–1.03×
updateLeaves, singleton | 38.49×
updateProof, 1,024 updates | 1.37–42.27×
multiproofSize, counting only | 6.54–113.43×
Full multiproof decoding | 1.16–1.34×
RHP modifyLeaves | 1.91–3.61×
Full RHP diff-proof verification | 1.01–1.73×

The 38x improvement in updateLeaves is due to allocations going from 1->0. Won't be as pronounced in practice.
The 113x multiproofSize improvement is for a very sparse tree, unlikely to occur to in practice.

RHP will probably not be meaningfully faster. Sync could be a bit faster, but sync is mainly bounded by disk, not CPU.

consensus/merkle.go
```diff
@@ -267,114 +277,83 @@ func (acc *ElementAccumulator) addLeaves(leaves []elementLeaf) [64][]types.Hash2
...
-			// Merge with the existing tree at this height. Since we're always
-			// adding leaves on the right-hand side of the tree, the existing
-			// root is always the left-hand sibling.
 			h = blake2b.SumPair(oldRoot, h)
 		}
 	}
+
+	// Every original tree that grew is a left sibling on the first added
+	// leaf's proof path. Above their merge point, they share the same proof.
+	for height, growth := range treeGrowth[:len(leaves[0].MerkleProof)] {
```
peterjan, line 303: `addLeaves` never clears the incoming MerkleProof, so is it possible for an attacker to land a spent ephemeral output at added[0] and shift every treeGrowth extension or cause a panic here?

lukechampine: good catch! I thought this would not be an issue because non-empty MerkleProofs would be caught during validation, but apparently that is not the case. We should add a `len(e.MerkleProof) == 0` check for ephemeral outputs. That would be a fork, though.

peterjan: Yeah I just had a feeling

